PR #7, opened 2019-01-31
add more cluster-config resources

deads2k:
Adds namespaces for config and infrastructures and networks.


deads2k:
required are green

PR #16, opened 2019-02-12
create bootstrap-manifests directory

deads2k:
this may solve the issue from https://github.com/openshift/installer/pull/1187

/assign @mfojtik


sttts:
This is not enough.

the Dockerfile has to copy `manifests/*` to `/usr/share/bootkube/manifests/manifests`. Note the second "manifests".

`/usr/share/bootkube/manifests` should be better called `/usr/share/bootkube/templates`. Then it would be clearer.

deads2k:
isntall completed merging

PR #17, opened 2019-02-19
Add CRD validation schemata

sttts:


hack/update-crd-validation/main.go
```diff
@@ -0,0 +1,143 @@
...
+	codecs = serializer.NewCodecFactory(scheme)
+)
+
+func init() {
+	v1beta1.AddToScheme(scheme)
+}
+
+func main() {
+	// load existing manifests from manifests/ dir
+	existingManifests, err := crdsFromDirectory("manifests")
```
deads2k: I'm ok starting here.   Seems like we should just generate everything though

manifests/0000_05_config-operator_01_authentication.crd.yaml
```diff
@@ -13,6 +14,12 @@ spec:
   subresources:
     status: {}
   versions:
-    - name: v1
-      served: true
-      storage: true
+  - name: v1
+    served: true
+    storage: true
+status:
```
deads2k: I'm not clear on why we're doing this, but ok.

deads2k:
@sttts had to tweak the generator?  I can live with that, but maybe a separate commit to make more easily pickable?

/lgtm

damemi:
Do we still want to pursue https://github.com/kubernetes-sigs/controller-tools/pull/144? Could do everything from the makefile without a hack file and a dependency on the controller tools. I'm open to either

hack/update-crd-validation/main.go
sttts: what else?

sttts: We have to copy this mechanism to all repos that have CRDs.

sttts:
> @sttts had to tweak the generator? I can live with that, but maybe a separate commit to make more easily pickable?

No tweaking. I copy files where the generator expects them, and I only copy the validation part. So whatever the generator does outside, we don't care. @damemi's PR against the generator will fix this, so we can make use of more if we want to.

Makefile
```diff
@@ -31,3 +31,14 @@ $(call add-bindata,v3.11.0,./bindata/v3.11.0/...,bindata,v311_00_assets,pkg/oper
 clean:
 	$(RM) ./cluster-config-operator
 .PHONY: clean
+
+update-codegen:
```
soltysh: Don't you want to move that to library-go?

soltysh: @tnozicka fyi

Makefile
```diff
@@ -31,3 +31,14 @@ $(call add-bindata,v3.11.0,./bindata/v3.11.0/...,bindata,v311_00_assets,pkg/oper
...
+	go run ./vendor/github.com/openshift/library-go/cmd/crd-schema-gen/main.go --apis-dir vendor/github.com/openshift/api
+.PHONY: update-codegen
+
+verify-codegen: update-codegen
```
tnozicka: this should not depend on update

Makefile
tnozicka: please rename the target to `update-codegen-crd` and you can hook it into `update-codegen: update-codegen-crd` if you want so it doesn't conflict with standart library-go targets

manifests/0000_05_config-operator_01_authentication.crd.yaml
sttts: fixed. That were default values that yaml marshalling added. Removed them by not going via CRD golang types.

sttts:
image operator.

/retest

sttts:
aws flake.

/retest

deads2k:
nothing has merged, looks like tide doesn't see approved. merging

PR #19, opened 2019-02-27
Generate CRD schema

damemi:
This is an update to https://github.com/openshift/cluster-config-operator/pull/17 which uses the fork of controller-tools checked into library-go


sttts:
ci/prow/verify-deps is red

damemi:
covered by https://github.com/openshift/cluster-config-operator/pull/17

PR #20, opened 2019-02-28
Add validation for logout redirect for the console CRD

zherman0:
Add validation for logout redirect for the console CRD
    1. Validation in the CRD to test that well formed URLs are supplied
       a) https://www.redhat.com:8080/logout.html is valid
       b) http://www.redhat.com is invalid because it is insecure
       c) www.redhat.com is invalid


benjaminapetersen:
@spadgett fyi

related to 
- stub for console config validation https://github.com/openshift/origin/pull/22103
- console operator config validation
   - https://github.com/openshift/console-operator/pull/152
   - https://github.com/openshift/console-operator/pull/165

manifests/0000_05_config-operator_01_console.crd.yaml
```diff
@@ -16,3 +16,14 @@ spec:
...
     storage: true
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            authentication:
+              properties:
+                logoutRedirect:
+                  pattern: ^$|^((https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))$
```
benjaminapetersen: Testing this with a regex tester:

```
https://foobar                              // valid, probably should not be?
https://redhat.com/#foo             // valid, i dont think we care about fragments
https://redhat.com?red=hat       // valid, i dont think we care about query strings
```

already stated in prev comment:

```
https://www.redhat.com:8080/logout.html  // valid
http://www.redhat.com                                   // invalid
www.redhat.com is invalid                              // invalid
```

spadgett: > `https://foobar                              // valid, probably should not be?`

Is there a reason we would require a path? I'd think this is OK

zherman0: I wanted https://localhost to work

benjaminapetersen: @spadgett I was more commenting on the `.<something>`.  
`localhost` does seem a valid point.

benjaminapetersen:
`make verify` doesn't like the `pattern`:

```bash
--- manifests/0000_05_config-operator_01_console.crd.yaml	2019-03-08 16:36:01.000000000 +0000
+++ /go/src/github.com/openshift/cluster-config-operator/101316955/0000_05_config-operator_01_console.crd.yaml	2019-03-08 16:37:50.753525862 +0000
@@ -38,7 +38,6 @@
             authentication:
               properties:
                 logoutRedirect:
-                  pattern: ^$|^((https):\/\/?)[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|\/?))$
                   description: 'An optional, absolute URL to redirect web browsers
                     to after logging out of the console. If not specified, it will
                     redirect to the default login page. This is required when using
```

zherman0:
I need https://github.com/openshift/api/pull/249 to merge first before the verify test will pass.

deads2k:
this doesn't look right.  You should need to re-vendor openshift/api and regenerate, rigth?

zherman0:
We will need to bump the deps now that api/pr249 has merged for this to pass

benjaminapetersen:
@zherman0 
```bash
# usually works
make update-deps && make generate && make verify
# but I think the deps update is manual in this repo, probably something like
glide update --strip-vendor && make update-codegen && make verify
```

zherman0:
This update to the manifest was automatically done when the vendor apis was bumped and the pattern auto created.  Therefore, closing this ticket.

PR #21, opened 2019-03-01
Add Scheduler CRD

sjenning:
@deads2k @ravisantoshgudimetla @derekwaynecarr 

mirrors `Scheduler` type in https://github.com/openshift/api/blob/master/config/v1/types_scheduling.go

### DIFF
--- a/manifests/0000_05_config-operator_01_scheduling.crd.yaml
+++ b/manifests/0000_05_config-operator_01_scheduling.crd.yaml
@@ -1,0 +1,16 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: schedulers.config.openshift.io
+spec:
+  group: config.openshift.io
+  scope: Cluster
+  names:
+    kind: Scheduler
+    singular: scheduler
+    plural: schedulers
+    listKind: SchedulerList
+  versions:
+    - name: v1
+      served: true
+      storage: true


manifests/0000_05_config-operator_01_scheduling.crd.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: scheduling.config.openshift.io
```
deads2k: Why isn't your resource plural like all the others?

manifests/0000_05_config-operator_01_scheduling.crd.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  group: config.openshift.io
+  scope: Cluster
+  names:
+    kind: Scheduling
+    singular: scheduling
+    plural: scheduling
```
deads2k: doing this means that you'll need to special case yourself in the client generator like https://github.com/openshift/origin/pull/22126

manifests/0000_05_config-operator_01_scheduling.crd.yaml
sjenning: `schedulings`?

smarterclayton:
I'm ok with giving things good names, that's why we added singular.

Any other objections?

PR #23, opened 2019-03-06
Fix config/v1 NetworkStatus spec

sttts:


danwinship:
This (and by extension https://github.com/openshift/api/pull/237) seems wrong; we shouldn't be allowing `null` as a value here, we should be marking the fields `omitempty` (in which case the CRD doesn't need to be updated because it already doesn't require the array to exist or be non-empty)

sttts:
@danwinship am fine with that. Installer uses null though.

sttts:
@danwinship can you open pulls which are correct? My day is coming to an end here, and I won't be able to push that through.

sttts:
With omitempty every writer of those objects (i.e. repos which marshal using openshift/api types), must be updated.

PR #25, opened 2019-03-08
add cluster resource quota resources

mfojtik:
depends on https://github.com/openshift/origin/pull/22277

/hold

### DIFF
--- a/manifests/0000_03_quota-openshift_01_clusterresourcequota.crd.yaml
+++ b/manifests/0000_03_quota-openshift_01_clusterresourcequota.crd.yaml
@@ -1,0 +1,76 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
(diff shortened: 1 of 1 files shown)


mfojtik:
@deads2k looks like we won't be able to keep the group name?

```
Mar 08 13:15:07 ip-10-0-14-27 bootkube.sh[4537]: Failed to create apiextensions.k8s.io/v1beta1, Resource=customresourcedefinitions: CustomResourceDefinition.apiextensions.k8s.io "quota.openshift.io" is invalid: metadata.name: Invalid value: "quota.openshift.io": must be spec.names.plural+"."+spec.group
```

mfojtik:
@deads2k i was expecting this to explode?

deads2k:
looks plausible, though I suspect that we'll end up having to fix the validation issues.

/lgtm

mfojtik:
@deads2k validation fixes: https://github.com/openshift/api/pull/267

PR #31, opened 2019-04-09
crq: add status sub-resource

mfojtik:
Missing this means openshift controller manager can't update status for cluster resource quota 🤦‍♂️


mfojtik:
Ci is broken, this is needed to unstick the origin PR and I manually tested this in my cluster and it works.

Button merging now.

PR #15, opened 2019-02-12
add empty CRs

deads2k:
@bparees since you asked

I *think* this will avoid stomping.  The test-aws should fail if this clears values.

PR #34, opened 2019-04-11
add empty CRs

deads2k:
from https://github.com/openshift/cluster-config-operator/pull/15


deads2k:
unauathorized.  This may be workingish

/retest

deads2k:
The dockerfile changes and file moves actually came into master in https://github.com/openshift/cluster-config-operator/pull/36 .  https://github.com/openshift/cluster-config-operator/pull/43 reverts the four empty config resources.

Most likely cause is authentication set here: https://github.com/openshift/cluster-authentication-operator/blob/36e2370500a73e638ac5730955f0e287a29f0a85/pkg/operator2/starter.go#L41-L48 , which doesn't reconcile and then causes https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/configobservation/auth/auth_metadata.go to never set the `authConfig` stanza.

PR #35, opened 2019-04-11
add empty CRs

deads2k:
from #15


deads2k:
looks like this is the one triggering the machine config.  one of these four

PR #36, opened 2019-04-11
add empty CRs

deads2k:
from #15


sjenning:
@deads2k do we need to remove the creation the installer does?
https://github.com/openshift/installer/tree/master/pkg/asset/manifests

PR #44, opened 2019-04-12
Re add minus auth

deads2k:
authentication is the most likely reason that the auth config observer failed


enj:
#45 contains all of the resources.

PR #45, opened 2019-04-12
Add back all empty custom resources (including authentication)

enj:
Signed-off-by: Monis Khan <mkhan@redhat.com>

Requires https://github.com/openshift/cluster-kube-apiserver-operator/pull/399

@openshift/sig-auth @openshift/sig-master


deads2k:
/lgtm
/hold

holding for prereq

enj:
/retest

I cannot tell if the KCM crash loop is related.

PR #50, opened 2019-04-29
monitoring: add servicemonitor CRD

mfojtik:
Each control plane operator that run early in bootstrap process is installing a ServiceMonitor, so prometheus can scrape metrics from it.

Right now, the prometheus operator is installing this CRD, which cause transient, but noisy errors where operator is retrying to create ServiceMonitor instance until prometheus operator is installed.

As this CRD is common enough to come using cluster config operator, we should install it with rest of the bootstrap manifests.

/cc @openshift/sig-monitoring 
/cc @deads2k

### DIFF
--- a/manifests/0000_05_config-operator_10_servicemonitor.crd.yaml
+++ b/manifests/0000_05_config-operator_10_servicemonitor.crd.yaml
@@ -1,0 +1,292 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: servicemonitors.monitoring.coreos.com
+spec:
+  group: monitoring.coreos.com
+  names:
+    kind: ServiceMonitor
+    plural: servicemonitors
+  scope: Namespaced
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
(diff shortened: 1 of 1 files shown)


mfojtik:
/hold

We need to coordinate this with removing the CRD in prometheus operator (otherwise CVO will stomp their). That means we should freeze the prometheus operator CRD for the time this land and is removed in prometheus operator.

brancz:
I understand and agree with the general reasoning for this, but does this have to land for 4.1? I'd prefer not to make unnecessary changes, and is it somehow possible for the cluster-monitoring-operator to put these CRD definition in its payload instead of manually having to update this CRD? It feels error prone to have to remember to sync it here.

squat:
It could certainly go in the CMO's own manifests just as it is placed in the CCO's manifests here, but that would not significantly reduce the time that containers early in the installation pipeline see these transient errors. Nevertheless, I agree that updates to the CRDs have the potential to get delayed here.

deads2k:
> Each control plane operator that run early in bootstrap process is installing a ServiceMonitor, so prometheus can scrape metrics from it.
> 
> Right now, the prometheus operator is installing this CRD, which cause transient, but noisy errors where operator is retrying to create ServiceMonitor instance until prometheus operator is installed.

I think that it's proper for the cluster-monitoring-operator to create these.   They are not required to bootstrap the first set of operators.  Everything else in this repo is about being able to create the first operators to start the control plane.

/close

PR #51, opened 2019-04-29
 manually create a CRD that should be generated

deads2k:

PR #32, opened 2019-04-09
Add RoleBindingRestriction CRD

sallyom:
Moving RoleBindingRestrictions to CRD, using @mfojtik 's example with ClusterResourceQuota
Bug 1684344

related:
https://github.com/openshift/origin/pull/22534
https://github.com/openshift/api/pull/283

PR #53, opened 2019-04-30
Add rolebindingrestriction and securitycontextconstraint CRDs

damemi:
Stealing https://github.com/openshift/cluster-config-operator/pull/51, this generates the CRDs instead of manually creating them

For reference, my process for generating the CRDs is a little hacky mainly due to some minor limitations and missing features in the [controller-tools generator](https://github.com/kubernetes-sigs/controller-tools/tree/master/cmd/crd) and our [fork](https://github.com/openshift/kubernetes-sigs-controller-tools). Here's the process I used today for posterity:

1. Build the [crd tool from the origin-4.1-kubernetes-1.13.4 branch of our fork](https://github.com/openshift/kubernetes-sigs-controller-tools/tree/origin-4.1-kubernetes-1.13.4/cmd/crd). This branch is currently stably rebased from the upstream tool, but our fork fundamentally handles optional/required fields opposite of upstream.
2. Run `crd generate --domain openshift.io --apis-path vendor/github.com/openshift/api/authorization/v1 --apis-pkg github.com/openshift/api/authorization/v1` and `crd generate --domain openshift.io --apis-path vendor/github.com/openshift/api/security/v1 --apis-pkg github.com/openshift/api/security/v1` to generate the CRDs in a `config/crds` directory
3. Manually modify these 2 CRDs to add the following changes:
   * Remove `creationTimestamp: null`
   * Add `singular` and `listKind` names (singular is available with the `+kubebuilder:singular=foo` tag but wasn't set for either of these types in openshift/api. `listKind` isn't available as a tag but we can add it or may want to consider refactoring name tags)
   * Add `subresources` section (this is available as a `+kubebuilder:subresource:status` tag)
   * Add `versions` list, which I also don't think is available in kubebuilder
   * Remove `status` object
   * Change `scope` to `Cluster` for RBR (`+genclient:nonNamespaced` is available to generate this and was properly set for SCC in openshift/api)

Also the full controller-tools `crd` command requires a `repo` value to be set in a PROJECT file, like so:
```
domain: openshift.io
repo: ./
```
We tried to make the PROJECT file unnecessary in our fork but apparently we still lack the ability to pass that `repo` field as a flag. (todo)

As you can see, this still requires some manual work and we should look into filling out some of the crd builder tool's missing features


damemi:
For the missing tags I've opened https://github.com/openshift/api/pull/302

sallyom:
@damemi @deads2k 
I'm confused, this is same as this: https://github.com/openshift/cluster-config-operator/pull/32/commits/659758d2cd69536c776db43ffbe64911ad44d03d 

To generate, all I did was add the first 18 lines: https://github.com/openshift/cluster-config-operator/blob/659758d2cd69536c776db43ffbe64911ad44d03d/manifests/0000_03_authorization-openshift_01_rolebindingrestrictions.crd.yaml#L1-#L18    then ran `make update-codegen-crds`  
:woman_shrugging: 
Which of the 3 PRs that add this file are we merging?? (i don't really care, just want these changes to be in asap)

damemi:
This is old, closing

PR #54, opened 2019-05-01
remove all validation

deads2k:
It can come back later if it gets fixed.  The definitions were wrong.


deads2k:
I am willing to merge over verify if necessary.

PR #56, opened 2019-05-01
TEST PR DO NOT MERGE

sallyom:
/hold


deads2k:
This has tested green, but verify is ugly/borken.  We'll fix that later.  This should deflake us and make progress.

PR #57, opened 2019-05-01
Add RoleBindingRestriction CRD

sallyom:
This validation is good, at least I was able to create a RBR in my test cluster that had the old CRD (before the revert)


deads2k:
/lgtm

let's see if get lucky

PR #59, opened 2019-05-02
RBRs are Namespaced, and more nullable, update CRD

sallyom:
@damemi  I don't think I'll need to make any corresponding api change.  I don't see any namespace comment in openshift/api, I see nonNamespace, though.   
@deads2k thanks for the tip.

### DIFF
--- a/manifests/0000_03_authorization-openshift_01_rolebindingrestriction.crd.yaml
+++ b/manifests/0000_03_authorization-openshift_01_rolebindingrestriction.crd.yaml
@@ -11,7 +11,7 @@ spec:
     singular: rolebindingrestriction
   subresources:
     status: {}
-  scope: Cluster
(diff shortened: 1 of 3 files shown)


deads2k:
/retest

this will never pass upgrade jobs

damemi:
@sallyom if it doesn't have `nonNamespaced`, the generator assumes it is namespaced.

Semi-related: I have a PR open to fix up some of these tags I noticed which could use a review if you have time: https://github.com/openshift/api/pull/302

deads2k:
/lgtm

upgrade will not pass. this is expected.  Will merge on green e2e-aws

sallyom:
closing in favor of [PR with proper bump](https://github.com/openshift/cluster-config-operator/pull/61)

PR #18, opened 2019-02-21
Add Proxy CRD

adambkaplan:

PR #66, opened 2019-06-06
Add proxy CRD manifest.

patrickdillon:
This adds a manifest to create a proxy CR for cluster-wide proxy configuration.

Jira: [CORS-1074](https://jira.coreos.com/browse/CORS-1074)
API PR: openshift/api#335
Supercedes: #18 

cc @abhinavdahiya @adambkaplan


patrickdillon:
/hold
I didn't realize how this works. It appears that the manifests are auto-generated with `make update-codegen`. So this PR depends on openshift/api#335

adambkaplan:
@patrickdillon correct. Once openshift/api#335 merges, you'll need to update the dependencies in openshift/client-go, and then finally update the dependencies here.

patrickdillon:
openshift/client-go#107 has been opened but has not yet merged. not sure how strictly this depends on that pr merging, because tests are passing.
/hold cancel

manifests/0000_05_config-operator_01_proxy.crd.yaml
```diff
@@ -0,0 +1,49 @@
...
+    plural: proxies
+    singular: proxy
+  scope: Cluster
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  validation:
+    openAPIV3Schema:
+      properties:
```
adambkaplan: @soltysh Will a description populate in `oc explain`?

manifests/0000_05_config-operator_01_proxy.crd.yaml
```diff
@@ -0,0 +1,49 @@
...
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
```
adambkaplan: @patrickdillon I don't see any description for the `status` field. I think you need to bump dependencies after the api and client-go PRs merge.

patrickdillon: thanks for clarifying. I will bump once openshift/client-go#107 merges

cgwalters:
So e.g. the MCO should watch this and ensure the host upgrades are using this proxy too?

abhinavdahiya:
https://github.com/openshift/client-go/pull/107 has merged.

danehans:
> So e.g. the MCO should watch this and ensure the host upgrades are using this proxy too?


@cgwalters when you have a moment, please review the cluster-wide proxy [design doc](https://docs.google.com/document/d/12bBF7GTgscW8B3apVU2WagtQpWh-PWk2tOWiD0dsdfU/edit?usp=sharing). The current plan is for operators to watch the `Proxy` api object and use the [status fields](https://github.com/openshift/api/blob/master/config/v1/types_proxy.go#L42-L55) to configure operands as needed. Feel free to comment in the doc with any questions or concerns.

manifests/0000_05_config-operator_01_proxy.crd.yaml
damemi: this is a known bug with the generator (top-level descriptions not being generated) that will hopefully be fixed soon when we switch to the new version of upstream controller-tools

manifests/0000_10_config-operator_01_proxy.crd.yaml
```diff
@@ -0,0 +1,69 @@
...
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  names:
+    kind: Proxy
+    listKind: ProxyList
+    plural: proxies
+    singular: proxy
+  validation:
```
deads2k: appears to be missing subresources, see https://github.com/openshift/cluster-config-operator/blob/master/manifests/0000_10_config-operator_01_image.crd.yaml#L17-L18 as an example.

PR #70, opened 2019-07-11
Use new crd-schema-gen in Makefile

damemi:
This uses the new CRD schema gen and should fix bugs around descriptions not being shown in oc explain


damemi:
Trying to debug ci failure, found this in kube-apiserver-operator logs:
```
type: 'Warning' reason: 'ServiceMonitorCreateFailed' Failed to create ServiceMonitor.monitoring.coreos.com/v1: the server could not find the requested resource
W0715 18:37:00.541769       1 observe_images.go:134] image.config.openshift.io/cluster: not found
W0715 18:37:00.541816       1 auth_metadata.go:44] authentications.config.openshift.io/cluster: not found
W0715 18:37:00.541861       1 observe_images.go:84] image.config.openshift.io/cluster: not found
W0715 18:37:00.541919       1 observe_scheduler.go:33] scheduler.config.openshift.io/cluster: not found
W0715 18:37:00.543622       1 observe_images.go:40] image.config.openshift.io/cluster: not found
W0715 18:37:03.626052       1 auth_metadata.go:44] authentications.config.openshift.io/cluster: not found
W0715 18:37:03.626078       1 observe_scheduler.go:33] scheduler.config.openshift.io/cluster: not found
W0715 18:37:03.626087       1 observe_images.go:84] image.config.openshift.io/cluster: not found
W0715 18:37:03.626101       1 observe_images.go:134] image.config.openshift.io/cluster: not found
W0715 18:37:03.626109       1 observe_images.go:40] image.config.openshift.io/cluster: not found
```
looks like these resources aren't found, possibly an error in their validations. @deads2k you pointed this out to me, do you have any idea where to look to see what tries to create these resources?

damemi:
Made some changes:
1. Bumped to bring in missing `// +optional` marker in API
2. Removed apiextensions-apiserver dep
3. Updated Makefile to generate CRDs for *all* manifest (not just the `config` types, we also have `quota`, `authorization`, and `security` manifests). I was hoping I could do this with one line that just passes all of `openshift/api` to the generator but this caused the generator to fail with a lot of errors like this: `github.com/openshift/cluster-config-operator/230188424/pkg/apis/config/v1:-: CRD for Network.config.openshift.io has no storage version`. Maybe something to look into but I don't think this adds too many operations since we only clone the generator once

damemi:
And it seems bumping has caused prometheus to fail on the procfs dependency of... itself? Maybe something with pinned versions of deps pinning versions of other deps

manifests/0000_10_config-operator_01_authentication.crd.yaml
```diff
@@ -79,9 +452,15 @@ spec:
...
                         type: string
+                    required:
+                    - name
                     type: object
+                required:
+                - kubeConfig
                 type: object
               type: array
+          required:
+          - type
```
sttts: @damemi why is this required if it has no required tag in https://github.com/openshift/api/blob/master/config/v1/types_authentication.go#L29 ?

PR #72, opened 2019-07-15
Add empty Proxy resource

sttts:


deads2k:
/approve

we should make a generator for these.

PR #74, opened 2019-07-23
Bump openshift/api and update CRD descriptions

sttts:
Follow-up of https://github.com/openshift/api/pull/386


spadgett:
Any chance you can include https://github.com/openshift/api/pull/387 as well? :)

PR #75, opened 2019-07-23
[release-4.1] Bug 1732934: remove defaultservingcert description in apiserver config

damemi:
Backporting to 4.1 for https://bugzilla.redhat.com/show_bug.cgi?id=1731105, this was fixed in master in https://github.com/openshift/cluster-config-operator/pull/70


cmd/cluster-config-operator/main.go
```diff
@@ -10,8 +10,8 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
-	utilflag "k8s.io/apiserver/pkg/util/flag"
-	"k8s.io/apiserver/pkg/util/logs"
+	utilflag "k8s.io/component-base/cli/flag"
```
deads2k: I didn't  expect this to change

glide.yaml
```diff
@@ -2,6 +2,8 @@ package: github.com/openshift/cluster-config-operator
 import:
 - package: github.com/openshift/library-go
   version: master
```
deads2k: release-4.1 i think

damemi: adding release-4.1 fixed the reason for needing to change the k8s imports above, updated thanks

deads2k:
you need to pin all the other dependencies. 

This is what we used to do in origin.
```
make WHAT=tools/depcheck/
_output/local/bin/linux/amd64/depcheck pin --glide.yaml=glide.yaml --glide.lock=glide.lock
```

jwforres:
I see two sets of API changes in this backport, the one referenced in the BZ, which is the defaultservingcert in the apiserver CRD.

However I also see changes to the infrastructure CRD. Where is the associated BZ for this change?  Who is the owner for this particular CRD?

damemi:
@jwforres you're right, though it seems odd that those changes to infrastructure in release-4.1 weren't previously carried to the release-4.1 branch of CCO. I think this may be a bug, and have opened https://bugzilla.redhat.com/show_bug.cgi?id=1734512 to track it (and additionally https://github.com/openshift/cluster-config-operator/pull/82 to fix it)

@wking it looks like you contributed the relevant changes in this PR to the infrastructure config API in https://github.com/openshift/api/pull/300, could you possibly comment on this as well?

deads2k:
/lgtm

these changes look correct.  Pinning to a particular SHA for one of our repos is strictly worse than pinning to the running branch.

wking:
> @wking it looks like you contributed the relevant changes in this PR to the infrastructure config API in openshift/api#300, could you possibly comment on this as well?

@damemi, are you asking for a 4.1.z backport of openshift/installer#1930?  Or did you want me to comment on something else?

damemi:
@wking ah, I see, if these config options are used by the installer then we can't have them in 4.1 without the 4.1 installer also supporting them. Unfortunately, I don't see a way we can get this kubeapiserver fix (which we do want backported) without also getting the infrastructure changes. If that's the case then I think we will need a backport for the installer as well.

Still, it's weird that these installer-related API changes made it into release-4.1 of the API but not the installer itself (or other components, as here). That seems like something we wouldn't want, right?

wking:
> if these config options are used by the installer then we can't have them in 4.1 without the 4.1 installer also supporting them. Unfortunately, I don't see a way we can get this kubeapiserver fix (which we do want backported) without also getting the infrastructure changes. If that's the case then I think we will need a backport for the installer as well.

Populating these fields for new 4.1 installs using some future version (e.g. 4.1.10) is possible.  The issue is migrating 4.1.0 and other early clusters (where they were not populated) to fill the new fields in if folks want to be able to rely on them being populated.  I'm not clear on how various consuming operators are being implemented; ideally they are looking for the new infrastructure fields and then falling back to the old, deprecated properties if the new fields are unset.  But before you can ignore the deprecated properties, we'll need migration tooling to port the data from the old locations to the new locations.  We have plans to manage that migration with [the config operator][1], but it doesn't have a migration-operator deployed in-cluster now (in either master or 4.1.z), so it would be a bit of a lift to add it.

Can you go into more detail about the Kubernetes API-server backport?  This particular PR has passed all its CI, and I guess fixes [the `explain` issue][2].  That isn't tied to installer/migration changes, is it?  Are there other changes you're interested in backporting that are?

[1]: https://github.com/openshift/cluster-config-operator
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1732934

damemi:
@wking Right, this fixes the `oc explain` issue I linked above. To fix that issue requires bumping the openshift/api dependencies here to the `release-4.1` branch in order to pull in the previously cherry-picked apiserver struct change (https://github.com/openshift/api/pull/375)

However prior to *that* cherry-pick going into 4.1, also merged were the changes to the infrastructure types I've pinged you on here (which I believe were just merged to master regularly before release-4.1 was cut, right?). Unfortunately those changes weren't propagated here to CCO in time, and now in order to get the changes I do want to backport also includes your changes.

So yes, the intent of this PR isn't tied to installer/migration changes but those infrastructure types have been caught in the crossfire and may need to be included as collateral changes here. As you mentioned this change has passed CI, so it looks clean, but I would appreciate your review/approval on the infrastructure changes (see https://github.com/openshift/cluster-config-operator/pull/75/commits/c79d1546a0163fc739c004bcd02d81179c96623a) and also to make you aware of them being included.

wking:
> As you mentioned this change has passed CI, so it looks clean, but I would appreciate your review/approval on the infrastructure changes (see c79d154) and also to make you aware of them being included.

I don't see a problem with documenting the new fields in 4.1.z, even if [the deprecation recommendation][1] is a bit early without the migration logic in place.  The new fields [are not populated][2], but that's not a problem as long as folks continue to fall-back on [`cluster-config-v1`][3].  So I don't think we need to wait on installer support or migration logic before landing this doc change.

However, if other folks feel that the doc change is too misleading, I'm fine landing a 4.1.z API patch that adjusts the deprecation wording to mention the possible `cluster-config-v1` fallback for the life of 4.1.z.  And I'm personally fine with holding this doc bump until we have time to backport the installer changes to 4.1.z (easy) and create and backport a migration operator (more work, but we'll have to do the creation work eventually anyway), but that's really up to the z-stream patch managers.  Am I missing something?

[1]: https://github.com/openshift/cluster-config-operator/commit/c79d1546a0163fc739c004bcd02d81179c96623a#diff-882e58d7c033dba57c8ae642018b7986R77
[2]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-config-operator/75/pull-ci-openshift-cluster-config-operator-release-4.1-e2e-aws/6/artifacts/e2e-aws/must-gather/cluster-scoped-resources/config.openshift.io/infrastructures/cluster.yaml
[3]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-config-operator/75/pull-ci-openshift-cluster-config-operator-release-4.1-e2e-aws/6/artifacts/e2e-aws/must-gather/namespaces/kube-system/core/configmaps.yaml

damemi:
@wking thanks, given that this is ultimately just a documentation change I don't see this creating much of a problem either.

@jwforres with that explanation of the infrastructure changes you mentioned, is this good to sign-off?

derekwaynecarr:
I think this is fine for 4.1.z.

Thanks for the explanation.

PR #78, opened 2019-07-25
Regenerate CRDs

damemi:
This pulls in changes from https://github.com/openshift/api/pull/390


manifests/0000_03_config-operator_01_proxy.crd.yaml
```diff
@@ -442,8 +442,6 @@ spec:
                   description: name is the metadata.name of the referenced config
                     map
                   type: string
-              required:
```
smarterclayton: This seems like it would be required.

manifests/0000_10_config-operator_01_authentication.crd.yaml
```diff
@@ -485,11 +477,7 @@ spec:
                   description: name is the metadata.name of the referenced config
                     map
                   type: string
-              required:
```
smarterclayton: Would expect this to be required

sttts: @smarterclayton wouldn't block the PR on that. It fixes too many other things. Opened https://github.com/openshift/api/pull/394 and we will follow-up to repair the required here.

manifests/0000_03_config-operator_01_proxy.crd.yaml
sttts: as above

manifests/0000_10_config-operator_01_scheduler.crd.yaml
```diff
@@ -441,8 +441,6 @@ spec:
                   description: name is the metadata.name of the referenced config
                     map
                   type: string
-              required:
-              - name
```
ravisantoshgudimetla: This LGTM

sttts:
@damemi sorry for merging early. Can you prepare a follow-up to clean up the manual bump?

PR #79, opened 2019-07-25
Regenerate CRDs to set auth spec "type" optional

damemi:
Pulls in changes from https://github.com/openshift/api/pull/392


damemi:
/hold
tmp bump still in place

PR #80, opened 2019-07-29
Revert fork dep from PR#78

damemi:
#78 got merged with a fork dep in place. This switches back to the actual api and runs update-codegen to pull in changes


manifests/0000_03_config-operator_01_proxy.crd.yaml
```diff
@@ -442,6 +442,8 @@ spec:
                   description: name is the metadata.name of the referenced config
                     map
                   type: string
+              required:
+              - name
```
damemi: These are correct, this field was manually specified required https://github.com/openshift/api/blob/master/config/v1/types.go#L20-L22

PR #82, opened 2019-07-30
[release-4.1] Bug 1734512: missing config updates for infrastructure crd

damemi:
Related to https://github.com/openshift/cluster-config-operator/pull/75, I think it may be a bug that these config changes exist in `release-4.1` of the OpenShift API but were not propagated to 4.1 of the config operator. Holding while I look more into this

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1734512


glide.yaml
```diff
@@ -2,9 +2,115 @@ package: github.com/openshift/cluster-config-operator
 import:
 - package: github.com/openshift/library-go
   version: master
+- package: github.com/openshift/api
+  version: dedfb47b1ff14bbd03727fa5720f63d8dc7bea5b
```
damemi: This is the exact commit where these changes were added. If I go any further (for example, pinning to `release-4.1` explicitly) it will also pull in the apiserver changes in https://github.com/openshift/cluster-config-operator/pull/75

deads2k:
pinning to a sha looks wrong.  #75 appears to fix two bugzillas.  go you

PR #88, opened 2019-08-23
Bump to golang 1.12 and update Makefile image target

damemi:
Need review on the image target, note that the name changes to match the examples in https://github.com/openshift/cluster-config-operator/pull/88

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
 # $2 - Dockerfile path
 # $3 - context directory for image build
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-$(GO_PACKAGE),./Dockerfile,.)
+$(call build-image,ocp-cluster-config-operator,registry.svc.ci.openshift.org/ocp/4.2:cluster-config-operator,./Dockerfile,.)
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
 # $2 - Dockerfile path
 # $3 - context directory for image build
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-$(GO_PACKAGE),./Dockerfile,.)
+$(call build-image,ocp-cluster-config-operator,registry.svc.ci.openshift.org/ocp/4.2:cluster-config-operator,./Dockerfile,.)
```
damemi: /cc @tnozicka 
is this correct? Changing the `origin-xxx` to `ocp-xxx` as the examples in https://github.com/openshift/library-go/pull/496

tnozicka: You should also change "FROM :4.0" in the rhel images and drop the origin one as there is no OKD yet. 

Might worth to do a subfolder for images but I'll leave it up on you.

I think https://github.com/openshift/oc is pretty updated as something to base it on but otherwise it looks good

damemi:
/hold
working on makefile changes

deads2k:
/approve

we need the correct golang levels to get generated resources correct.

PR #91, opened 2019-09-25
Bug 1755125: 4.2 - ClusterResourceQuota: allow integer values

sttts:

PR #92, opened 2019-09-25
ClusterResourceQuota: allow integer values

sttts:
Like https://github.com/openshift/cluster-config-operator/pull/91, but for master.

Working around https://github.com/kubernetes-sigs/controller-tools/issues/328.

cc @damemi – this is another patch we need to port when using the controller-tools generator.


sttts:
"dial tcp 10.130.0.48:8443: connect: No route to host" looks like network flake.

/retest

PR #93, opened 2019-09-26
Bug 1754640: Update CRD manifests for structural schema requirements

damemi:
This updates the config manifests using the new library-go crd-schema-gen. Works at least until https://github.com/openshift/api/pull/443 is sorted out.

/hold
depends on https://github.com/openshift/library-go/pull/532


damemi:
/hold cancel
this now has the library-go fix included

damemi:
I think this is resulting in unexpected behavior because we have `ClusterResourceQuota` defined twice:
1. https://github.com/openshift/cluster-config-operator/blob/master/manifests/0000_10_quota-openshift_01_clusterresourcequota.crd.yaml
2. https://github.com/openshift/cluster-config-operator/blob/master/manifests/0000_03_quota-openshift_01_clusterresourcequota.crd.yaml

Is there any reason for this? Should we delete one of them and, if so, which one? They look to be the same resource.

sttts:
> Is there any reason for this? Should we delete one of them and, if so, which one? They look to be the same resource.

Would keep the earlier one.

PR #94, opened 2019-09-30
bump(*)

enj:
Signed-off-by: Monis Khan <mkhan@redhat.com>

To pick up https://github.com/openshift/api/pull/429


glide.yaml
```diff
@@ -1,10 +1,9 @@
...
-  version: 505eaef017263e299324067d40ca2c48f6a2cf50
-- package: github.com/prometheus/procfs
-  version: 65c1f6f8f0fc1e2185eb9863a3bc751496404259
+  - package: k8s.io/klog
+    version: v1.0.0
+
+  - package: github.com/openshift/api
+    version: master
+  - package: github.com/openshift/library-go
+    version: master
```
sttts: have we bumped to kube 1.16?

manifests/0000_10_config-operator_01_network.crd.yaml
```diff
@@ -487,6 +327,7 @@ spec:
                   hostPrefix:
                     description: The size (prefix) of block to allocate to each node.
                     format: int32
+                    minimum: 0
```
sttts: cc @squeed I guess this is intentional.

glide.yaml
sttts: looks like we did seeing the schema changes.

PR #95, opened 2019-09-30
Migrate console extensions CRDs from console-operator

benjaminapetersen:
See https://github.com/openshift/console-operator/pull/302 for associated removal PR.  These files have existed in the console-operator, but should live here.

This is not a 1:1, however, due to `update-codegen-crds` modifying the files.

@spadgett @rhamilto @jhadvig fyi.

The `codegen` command overwrites these files based on what is in `openshift/api`, so what we previously had documented may no longer be here (if it was also not in `openshift/api`)

### DIFF
--- a/manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
+++ b/manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
@@ -1,0 +1,87 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: consoleclidownloads.console.openshift.io
+  annotations:
+    displayName: ConsoleCLIDownload
+    description: Extension for configuring openshift web console command line interface
+      (CLI) downloads.
+spec:
+  scope: Cluster
+  group: console.openshift.io
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  names:
+    plural: consoleclidownloads
+    singular: consoleclidownload
+    kind: ConsoleCLIDownload
+    listKind: ConsoleCLIDownloadList
+  additionalPrinterColumns:
+  - name: Display name
+    type: string
+    JSONPath: .spec.displayName
+  - name: Age
+    type: string
+    JSONPath: .metadata.creationTimestamp
+  - name: Description
+    type: string
+    JSONPath: .spec.description
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: ConsoleCLIDownload is an extension for configuring openshift web
+        console command line interface (CLI) downloads.
+      type: object
+      required:
+      - spec
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          description: Standard object's metadata.
+          type: object
+        spec:
+          description: ConsoleCLIDownloadSpec is the desired cli download configuration.
+          type: object
+          required:
+          - description
+          - displayName
+          - links
+          properties:
+            description:
+              description: description is the description of the CLI download (can
+                include markdown).
+              type: string
+            displayName:
+              description: displayName is the display name of the CLI download.
+              type: string
+            links:
+              description: links is a list of objects that provide CLI download link
+                details.
+              type: array
+              items:
+                description: Represents a standard link that could be generated in
+                  HTML
+                type: object
+                required:
+                - href
+                - text
+                properties:
+                  href:
+                    description: href is the absolute secure URL for the link (must
+                      use https)
+                    type: string
+                  text:
+                    description: text is the display text for the link
+                    type: string
--- a/manifests/0000_10_config-operator_01_consoleexternalloglink.crd.yaml
+++ b/manifests/0000_10_config-operator_01_consoleexternalloglink.crd.yaml
@@ -1,0 +1,85 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: consoleexternalloglinks.console.openshift.io
+  annotations:
+    displayName: ConsoleExternalLogLinks
+    description: ConsoleExternalLogLink is an extension for customizing OpenShift
+      web console log links.
+spec:
+  scope: Cluster
+  group: console.openshift.io
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  names:
+    plural: consoleexternalloglinks
+    singular: consoleexternalloglink
+    kind: ConsoleExternalLogLink
+    listKind: ConsoleExternalLogLinkList
+  additionalPrinterColumns:
+  - name: Text
+    type: string
+    JSONPath: .spec.text
+  - name: HrefTemplate
+    type: string
+    JSONPath: .spec.hrefTemplate
+  - name: Age
+    type: date
+    JSONPath: .metadata.creationTimestamp
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: ConsoleExternalLogLink is an extension for customizing OpenShift
+        web console log links.
+      type: object
+      required:
+      - spec
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          description: Standard object's metadata.
+          type: object
+        spec:
+          description: ConsoleExternalLogLinkSpec is the desired log link configuration.
+            The log link will appear on the logs tab of the pod details page.
+          type: object
+          required:
+          - hrefTemplate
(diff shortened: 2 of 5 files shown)


sttts:
> Was speaking to @mfojtik regarding this migration and was told that these CRDs probably should stay in the console-operator repo, since cluster-config-operator should contain CRDs that are installed during the bootstrap/installation

Bootstrapping and installation is not the point here. All resources which are needed by other operators (especially those which come first in the installation order), must go to this repo. I understood that this is the case for the console types: kube-apiserver operator might want to install an extension for the console. It can only do that if the CRD is already installed in the system.

mfojtik:
Agree with @sttts. I thought these are just console specific CRD's to support extensions.

/lgtm

sttts:
Needs a rebase and CRD regeneration (we moved to 1.16).

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
```diff
@@ -0,0 +1,295 @@
...
+              type: string
+            uid:
+              description: "UID is the unique in time and space value for this object.
+                It is typically generated by the server on successful creation of
+                a resource and is not allowed to change on PUT operations. \n Populated
+                by the system. Read-only. More info: http://kubernetes.io/docs/user-guide/identifiers#uids"
+              type: string
+          type: object
+        spec:
+          description: ConsoleCLIDownloadSpec is the desired cli download configuration.
```
benjaminapetersen: Looks correct.

https://github.com/openshift/api/blob/master/console/v1/types_console_cli_download.go#L18

manifests/0000_10_config-operator_01_consoleexternalloglink.crd.yaml
```diff
@@ -0,0 +1,293 @@
...
+                in 1.21 release."
+              type: string
+            uid:
+              description: "UID is the unique in time and space value for this object.
+                It is typically generated by the server on successful creation of
+                a resource and is not allowed to change on PUT operations. \n Populated
+                by the system. Read-only. More info: http://kubernetes.io/docs/user-guide/identifiers#uids"
+              type: string
+          type: object
+        spec:
```
benjaminapetersen: also correct
https://github.com/openshift/api/blob/master/console/v1/types_console_external_log_links.go#L19

manifests/0000_10_config-operator_01_consolelink.crd.yaml
```diff
@@ -0,0 +1,315 @@
...
+                in 1.21 release."
+              type: string
+            uid:
+              description: "UID is the unique in time and space value for this object.
+                It is typically generated by the server on successful creation of
+                a resource and is not allowed to change on PUT operations. \n Populated
+                by the system. Read-only. More info: http://kubernetes.io/docs/user-guide/identifiers#uids"
+              type: string
+          type: object
+        spec:
```
benjaminapetersen: https://github.com/openshift/api/blob/master/console/v1/types_console_link.go#L18

manifests/0000_10_config-operator_01_consolenotification.crd.yaml
```diff
@@ -0,0 +1,295 @@
...
+                in 1.21 release."
+              type: string
+            uid:
+              description: "UID is the unique in time and space value for this object.
+                It is typically generated by the server on successful creation of
+                a resource and is not allowed to change on PUT operations. \n Populated
+                by the system. Read-only. More info: http://kubernetes.io/docs/user-guide/identifiers#uids"
+              type: string
+          type: object
+        spec:
```
benjaminapetersen: https://github.com/openshift/api/blob/master/console/v1/types_console_notification.go#L18

benjaminapetersen:
Rebased & regeneration checks out.  Should be good to go.

benjaminapetersen:
Bump for approval again, then I can move along the removal from `console-operator` [here](https://github.com/openshift/console-operator/pull/302).

benjaminapetersen:
Ah, I see another rebase needed, fixing quick.

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
```diff
@@ -0,0 +1,295 @@
...
+              type: string
+            links:
+              description: links is a list of objects that provide CLI download link
+                details.
+              items:
+                description: Represents a standard link that could be generated in
+                  HTML
+                properties:
+                  href:
+                    description: href is the absolute secure URL for the link (must
```
sttts: you can use `pattern: "<regex>"` to restrict the value.

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
```diff
@@ -0,0 +1,295 @@
...
+              type: string
+          type: object
+        spec:
+          description: ConsoleCLIDownloadSpec is the desired cli download configuration.
+          properties:
+            description:
+              description: description is the description of the CLI download (can
+                include markdown).
+              type: string
+            displayName:
```
sttts: can this be empty?

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
```diff
@@ -0,0 +1,295 @@
...
+                details.
+              items:
+                description: Represents a standard link that could be generated in
+                  HTML
+                properties:
+                  href:
+                    description: href is the absolute secure URL for the link (must
+                      use https)
+                    type: string
+                  text:
```
sttts: can this be empty?

manifests/0000_10_config-operator_01_consoleexternalloglink.crd.yaml
```diff
@@ -0,0 +1,293 @@
...
+                \n e.g., https://example.com/logs?resourceName=${resourceName}&containerName=${containerName}&resourceNamespace=${resourceNamespace}&podLabels=${podLabels}"
+              type: string
+            namespaceFilter:
+              description: namespaceFilter is a regular expression used to restrict
+                a log link to a matching set of namespaces (e.g., `^openshift-`).
+                The string is converted into a regular expression using the JavaScript
+                RegExp constructor. If not specified, links will be displayed for
+                all the namespaces.
+              type: string
+            text:
```
sttts: can this be empty?

manifests/0000_10_config-operator_01_consoleexternalloglink.crd.yaml
```diff
@@ -0,0 +1,293 @@
...
+                It is typically generated by the server on successful creation of
+                a resource and is not allowed to change on PUT operations. \n Populated
+                by the system. Read-only. More info: http://kubernetes.io/docs/user-guide/identifiers#uids"
+              type: string
+          type: object
+        spec:
+          description: ConsoleExternalLogLinkSpec is the desired log link configuration.
+            The log link will appear on the logs tab of the pod details page.
+          properties:
+            hrefTemplate:
```
sttts: use `pattern` to restrict it to https

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
```diff
@@ -0,0 +1,87 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
benjaminapetersen: Rebase, and I have to say I like the output of this much better.  It's limited to the fields that specifically matter to the custom resource.

sttts:
Validation as follow-up.

/lgtm

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
benjaminapetersen: It's a required field, so it should not be empty.

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
benjaminapetersen: Perhaps we assumed `required` would validate that it is not empty.  I'm guessing `present but empty` is allowed without regex, given your questions.  We should update.

spadgett:
Looks like we had a mismatch between openshift/api and what was in the CRD:

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_console/2822/pull-ci-openshift-console-master-e2e-aws-console/9623/artifacts/e2e-aws-console/gui_test_screenshots/9f3d29397d207d3c8bddfb64ad2c5ecd.png

Link text is now required, when it wasn't before. (Breaking the console merge queue)

spadgett:
Console queue fixed by https://github.com/openshift/console/pull/2888

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
benjaminapetersen: Are we allowed to change this now?  API has been in for a release already.

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
benjaminapetersen: Are we allowed to change this now?  API has been in for a release already.

manifests/0000_10_config-operator_01_consoleexternalloglink.crd.yaml
benjaminapetersen: Same, don't know if we can change it at this point.

manifests/0000_10_config-operator_01_consoleclidownload.crd.yaml
benjaminapetersen: This I need to fix as it has been part of the historical API.  We have been using:
```
pattern: '^https://([\w-]+.)+[\w-]+(/[\w- ./?%&=])?$'
```
Just need to figure out how to get this in `openshift/api` to generate.

benjaminapetersen: Ok this should be resolved via https://github.com/openshift/api/pull/457/files#r331225859 on our next regeneration.

manifests/0000_10_config-operator_01_consoleexternalloglink.crd.yaml
benjaminapetersen: https://github.com/openshift/api/pull/457/files#r331226280 should resolve on our next regeneration PR.

deads2k:
> See [openshift/console-operator#302](https://github.com/openshift/console-operator/pull/302) for associated removal PR. These files have existed in the console-operator, but should live here.
> 
> This is not a 1:1, however, due to `update-codegen-crds` modifying the files.
> 
> @spadgett @rhamilto @jhadvig fyi.
> 
> The `codegen` command overwrites these files based on what is in `openshift/api`, so what we previously had documented may no longer be here (if it was also not in `openshift/api`)

This should not have landed in this repo.  This is not about cross dependencies, this is about bootstrapping a cluster.  This is not a dumping ground for, "hey something else might/does use this".

PR #96, opened 2019-10-01
manifests: update apiserver config CRD with EncryptionType spec

sttts:
Follow-up of openshift/api#453.

glide.yaml had been broken in https://github.com/openshift/cluster-config-operator/pull/94.


glide.yaml
```diff
@@ -1,9 +1,15 @@
 package: github.com/openshift/cluster-config-operator
 import:
-  - package: k8s.io/klog
-    version: v1.0.0
+- package: k8s.io/klog
+  version: v1.0.0
```
mfojtik: @sttts can you bump this to version v0.4.0 ? (it is what origin use and I think 1.16.x)

sttts: done

enj:
LGTM, verify failure looks real.

PR #98, opened 2019-10-10
Fix local image build

tnozicka:
Fixes local image build failure with imagebuilder https://github.com/openshift/imagebuilder/issues/140

/cc @sallyom

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+vendor/github.com/prometheus/procfs/fixtures/


sallyom:
this is a bandaid, but it works - although have to work around [this](https://github.com/openshift/imagebuilder/issues/144), also :woman_shrugging:

soltysh:
> this is a bandaid, but it works - although have to work around [this](https://github.com/openshift/imagebuilder/issues/144), also woman_shrugging

Yeah, I've noticed that, but you can just pull that image and re-try, which is better than not being able to build at all :wink:

PR #99, opened 2019-10-11
Add ConsoleYAMLSample CRD as part of the console extensions

jhadvig:
Wasn't really sure what to put the the `apiVersion` and `kind` description, since it's "TBD" in @rhamilto [PR](https://github.com/openshift/console/pull/2889).  For now I've add `apiVersion` and `kind` description to the default we use in all of our CRDs.

@spadgett FYI

/assign @sttts


spadgett:
@jhadvig I believe this goes to openshift/api first and the CRD itself is generated

cc @benjaminapetersen

benjaminapetersen:
And just a note we will likely have to do something like this:
https://github.com/openshift/origin/pull/23382/files
to get tests to pass.

damemi:
The CRD generator will generate only the validation section (`make update-codegen`). As long as the API change has been bumped here (`make update-deps`), then just running the generator now should fill out the rest of the CRD

jhadvig:
@damemi PR updated. PTAL

damemi:
/hold
We're moving these CRDs to be consumed by this operator from openshift/api right now, and don't want to add more to the list until the current ones are already sorted out. See https://github.com/openshift/cluster-config-operator/pull/102

benjaminapetersen:
We may be able to `close` this one since the other will contain it + the rest.

benjaminapetersen:
/close

We will include this in 
https://github.com/openshift/console-operator/pull/314

PR #100, opened 2019-10-11
Makefile allow passing image tag to 'make images'

sallyom:
This allows passing a custom image registry to `make images`, like so:
`make IMAGE_REGISTRY=quay.io/sallyom images`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,13 +9,15 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
 	targets/openshift/crd-schema-gen.mk \
 )
 
+IMAGE_REGISTRY?=registry.svc.ci.openshift.org
+
 # This will call a macro called "build-image" which will generate image specific targets based on the parameters:
 # $0 - macro name
 # $1 - target suffix
 # $2 - Dockerfile path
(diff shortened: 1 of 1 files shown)


deads2k:
/approve

looks plausible.  Get @tnozicka to lgtm

/assign @tnozicka

Makefile
```diff
@@ -9,13 +9,14 @@ include $(addprefix ./vendor/github.com/openshift/library-go/alpha-build-machine
 	targets/openshift/crd-schema-gen.mk \
 )
 
+IMAGE_TAG?=registry.svc.ci.openshift.org/ocp/4.2:cluster-config-operator
```
soltysh: `IMAGE_TAG` might be confusing, since this is rather pull spec, also I'm not sure we want to allow redlining this other than just the repository part like in https://github.com/openshift/cluster-kube-scheduler-operator/tree/master/Makefile, for example.

soltysh: Also, add a blank line after this variable, like we do in other files, it improves readability.

sallyom: added, also updated to set `IMAGE_REGISTRY` instead of `IMAGE_TAG`

PR #101, opened 2019-10-14
Add add-crd-gen line to Makefile, regenerate console extension CRDs

benjaminapetersen:
Generating new code for CRDs from `openshift/api` doesn't seem to be working.  Working on a fix.
- Adds a line to `Makefile` : `$(call add-crd-gen,manifests,$(CRD_SCHEMA_GEN_APIS),./manifests,./manifests)`.
- Adds `vendor` after `make update-deps` as separate commit.
- Regenerates `console` `extension` CRDs as 3rd commit, which happens to also pickup `apiserver` and `infrastructure` changes. 

/cc @spadgett @bparees 
/assign @damemi


spadgett:
Do we want to wait for https://github.com/openshift/api/pull/465 ?

benjaminapetersen:
That or the next regen will pick it up.  Since this PR contains a fix in the `Makefile` I'm tempted to get it merged and do [openshift/api#465](https://github.com/openshift/api/pull/465) as a follow.

benjaminapetersen:
Regenerated to pick up 1 new file, looks like a lot came with it.

benjaminapetersen:
/retest

Timeouts.

Makefile
```diff
@@ -36,6 +36,8 @@ clean:
 CRD_SCHEMA_GEN_APIS := $(shell echo ./vendor/github.com/openshift/api/{authorization/v1,config/v1,quota/v1,security/v1,operator/v1alpha1,console/v1})
 CRD_SCHEMA_GEN_VERSION :=v0.2.1
 
+$(call add-crd-gen,manifests,$(CRD_SCHEMA_GEN_APIS),./manifests,./manifests)
```
soltysh: This should not be needed. With https://github.com/openshift/api/pull/470 in place and it looks like your bump already has that you should re-use the manifests provided by openshift/api. Sync with @damemi I thought he wanted to do it :wink:

damemi: Until https://github.com/openshift/cluster-config-operator/pull/102 merges those CRDs in openshift/api aren't picked up anywhere, meaning we still need to update here if you want changes pulled in

damemi:
/hold
We're moving these CRDs to be consumed by this operator from openshift/api right now, and don't want to add more to the list until the current ones are already sorted out. See https://github.com/openshift/cluster-config-operator/pull/102

PR #102, opened 2019-10-15
Copy manifests from openshift/api to builder

damemi:
This uses the CRDs which now live in openshift/api (https://github.com/openshift/api/pull/470) and removes those CRDs from this repo, as well as the codegen Makefile targets


Dockerfile
```diff
@@ -7,8 +7,19 @@ RUN GODEBUG=tls13=1 go build -ldflags "-X $GO_PACKAGE/pkg/version.versionFromGit
 FROM registry.svc.ci.openshift.org/ocp/4.2:base
 RUN mkdir -p /usr/share/bootkube/manifests/manifests
 RUN mkdir -p /usr/share/bootkube/manifests/bootstrap-manifests
-COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/manifests/* /usr/share/bootkube/manifests/manifests/
+COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/config/v1/*.yaml /usr/share/bootkube/manifests/manifests
+COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/quota/v1/*.yaml /usr/share/bootkube/manifests/manifests
+COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/security/v1/*.yaml /usr/share/bootkube/manifests/manifests
+COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/authorization/v1/*.yaml /usr/share/bootkube/manifests/manifests
+COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/console/v1/*.yaml /usr/share/bootkube/manifests/manifests
```
deads2k: wow, this looks wrong.  They need to fix this.

Dockerfile
```diff
@@ -7,8 +7,19 @@ RUN GODEBUG=tls13=1 go build -ldflags "-X $GO_PACKAGE/pkg/version.versionFromGit
...
+COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/operator/v1/*.yaml /usr/share/bootkube/manifests/manifests
```
deads2k: :(  This shouldn't be

PR #103, opened 2019-10-16
Revert "Migrate console extensions CRDs from console-operator"

benjaminapetersen:
Reverts openshift/cluster-config-operator#95

/hold
for https://github.com/openshift/console-operator/pull/314

/cc @deads2k @spadgett


deads2k:
/lgtm
/retest

/hold cancel

If this can pass CI, then it's good enough to be in master while you get them back at the right levels.

deads2k:
> /hold
> this will break console merge queue, can you give us just a little time to fix?

Sure.  Let's call it tomorrow AM.  Keeping these in a structural risk for people trying to use them an effectively pinning usage here. We've seen it happen with other CRDs and typos we've had.  At a certain point, we will have to correct the signal to the rest of product/payload even if some eggs break.

PR #105, opened 2019-10-25
bump(*)

damemi:
Bump openshift/api to pull in correct preserveUnknownFields values from https://github.com/openshift/api/pull/496


vendor/github.com/openshift/api/operator/v1/0000_70_console-operator.crd.yaml
```diff
@@ -0,0 +1,197 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
sttts: this new file is intentional?

sttts:
/approve

/hold
because of the question. Unhold at will.

vendor/github.com/openshift/api/operator/v1/0000_70_console-operator.crd.yaml
soltysh: This was added in the mean time to openshift/api so it will be here as well.
/hold cancel

damemi: We don't copy anything from operator/v1, and only one specific file from operator/v1alpha1 https://github.com/openshift/cluster-config-operator/blob/master/Dockerfile#L14

damemi:
/hold cancel

I think we answered your question?

PR #108, opened 2019-11-01
Replace kube_apiserver_operator with config_operator in build_info

benjaminapetersen:
Replace `kube_apiserver_operator` with `config_operator` in `build_info` metric.

Looks like a small copy/paste error.  If there is a reason its reporting an `apiserver` metric, however, I will close. 

/assign @deads2k @damemi

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -36,8 +36,8 @@ func Get() version.Info {
 func init() {
 	buildInfo := prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Name: "openshift_cluster_kube_apiserver_operator_build_info",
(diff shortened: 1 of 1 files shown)


damemi:
Looks like this is a copy paste error? https://github.com/search?q=org%3Aopenshift+openshift_cluster_kube_apiserver_operator_build_info&type=Code

Looking at similar metrics in other operators though, I also see them being re-used (like in oas-o and svcat: https://github.com/search?q=org%3Aopenshift+openshift_cluster_openshift_apiserver_operator_build_info&type=Code)

Is this intentional? I can't tell

damemi:
And 3 instances for service serving cert signer: https://github.com/search?q=org%3Aopenshift+openshift_service_serving_cert_signer_build_info&type=Code

PR #110, opened 2020-01-22
Bump API to include VSphere infrastructure

patrickdillon:
We need to pull in the changes from openshift/api#541 so that the `cluster-infrastructure-02-config.yml` manifest generated by the installer can be correctly applied. 

This commit was created by running `make update-deps`.

It seems extremely large so perhaps I am doing something very wrong. Guidance would be appreciated.


patrickdillon:
/close 
in favor of #111

PR #117, opened 2020-03-26
add operator manifests and run the operator

mfojtik:


manifests/0000_30_config-operator_07_deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+        command: ["cluster-config-operator", "operator"]
+        resources:
+          requests:
+            memory: 50Mi
+            cpu: 10m
+        volumeMounts:
+        env:
+        - name: IMAGE
+          value: quay.io/openshift/origin-cluster-config-operator:v4.0
+        - name: OPERATOR_IMAGE
```
deads2k: shouldn't be needed.

manifests/0000_30_config-operator_08_clusteroperator.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: config-operator
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec: {}
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
+    - name: openshift-apiserver
```
deads2k: no

deads2k:
largely ok with it.  If you go green, I'm ok merging and iterating.  If you don't, fix them up now

/lgtm
/hold

manifests/0000_30_config-operator_08_clusteroperator.yaml
mfojtik: good catch, bad copy&pasta

manifests/0000_30_config-operator_07_deployment.yaml
mfojtik: removed

manifests/0000_30_config-operator_08_clusteroperator.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: config-operator
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
+spec: {}
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
+    - name: config-operator
```
deads2k: I'm not sure this is going to be set.

deads2k:
I think we get 0000_10_

/lgtm

deads2k:
red looks permanent

/lgtm cancel

pkg/operator/starter.go
```diff
@@ -64,12 +67,39 @@ func RunOperator(ctx context.Context, controllerContext *controllercmd.Controlle
 
 	logLevelController := loglevel.NewClusterOperatorLoggingController(operatorClient, controllerContext.EventRecorder)
 
+	// As this operator does not manage any component/workload, report this operator as available and not progressing by default.
+	// TODO: Revisit this with full controller at some point.
+	operatorController := factory.New().ResyncEvery(10*time.Second).WithSync(func(ctx context.Context, controllerContext factory.SyncContext) error {
```
mfojtik: @deads2k i think we missing this bit in library-go... if you have generic operator (not static pod, not workloads), nothing really sets the `FooAvailable` anywhere, which means the status controller will union there as `"Unknown"` with `NoData` as reason... The CVO then freak out, as there is nothing that moves the operator Progressing to false and Available to true (it will stay "Unknown"). I'm 99% sure that is the reason why the e2e tests are failing.

Since this operator does not have any workload to manage and does not have any controller that would set "FooAvailable", I think it is reasonable to have something like below that "default" the operator conditions on startup (which is why I placed this to starter). WDYT?

deads2k: I prefer this as a separate controller in a separate package or you'll have people adding their, "just a small thing" here.

deads2k:
can be fixed up after.

/lgtm

PR #118, opened 2020-03-31
operator: add server to serve metrics and healthz

mfojtik:


manifests/0000_30_config-operator_04_servicemonitor.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
deads2k: duplicate?

manifests/0000_30_config-operator_04_servicemonitor.yaml
```diff
@@ -0,0 +1,60 @@
...
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-config
```
deads2k: what gets created in this namespace?

mfojtik: nothing prometheus worthy probably.

PR #119, opened 2020-03-31
aws_platform_service_location: add a controller that sync user specified service endpoints to status

abhinavdahiya:
The controller only works for AWS platform
The service endpoints are validated before syncing to the the status,
- no duplicates should be specified for a service
- the URL must be hostname or a valid URI without any request paths
If the service endpoints are not allowed or valid, the status service endpoints are not modified.

depends on https://github.com/openshift/api/pull/599


pkg/operator/aws_platform_service_location/controller.go
```diff
@@ -38,12 +49,94 @@ func NewController(operatorClient operatorv1helpers.OperatorClient, infraClient
...
+	if err != nil {
+		return err
+	}
+
+	currentInfra := obj.DeepCopy()
+	var platformName configv1.PlatformType
+	if pstatus := currentInfra.Status.PlatformStatus; pstatus != nil {
+		platformName = pstatus.Type
+	}
+	if platformName == "" {
```
deads2k, line 66: kube canonical is len() == 0

pkg/operator/aws_platform_service_location/controller.go
```diff
@@ -38,12 +49,94 @@ func NewController(operatorClient operatorv1helpers.OperatorClient, infraClient
...
 		return nil
 	}
-	klog.Info(spew.Sprintf("current state: %s", currentInfra.Status))
+	if err != nil {
+		return err
+	}
+
+	currentInfra := obj.DeepCopy()
+	var platformName configv1.PlatformType
+	if pstatus := currentInfra.Status.PlatformStatus; pstatus != nil {
```
deads2k, line 63: Seems like the `Status.PlatformStatus` and `Status.Platform` should always be the same.

pkg/operator/aws_platform_service_location/controller.go
```diff
@@ -38,12 +49,94 @@ func NewController(operatorClient operatorv1helpers.OperatorClient, infraClient
...
+	}
+
+	if !(currentInfra.Spec.PlatformSpec != nil &&
+		currentInfra.Spec.PlatformSpec.Type == configv1.AWSPlatformType &&
+		currentInfra.Spec.PlatformSpec.AWS != nil &&
+		len(currentInfra.Spec.PlatformSpec.AWS.ServiceEndpoints) > 0) {
+		return nil // nothing to do here
+	}
+
+	services := currentInfra.Spec.PlatformSpec.AWS.ServiceEndpoints
```
deads2k, line 82: deep copy please

deads2k:
needs owners file

pkg/operator/aws_platform_service_location/controller.go
abhinavdahiya: these should be same, but aren't currently. Migrations should make them same. and then we can drop the handling.

pkg/operator/aws_platform_service_location/controller.go
abhinavdahiya: https://github.com/openshift/cluster-config-operator/pull/119/commits/d957d968accfb543daecb6c1d7e10ecfd8584467#diff-51a3ce64d08707b029401c01f6a81ac0R61

already deepcopies.

deads2k: > [d957d96#diff-51a3ce64d08707b029401c01f6a81ac0R61](https://github.com/openshift/cluster-config-operator/commit/d957d968accfb543daecb6c1d7e10ecfd8584467#diff-51a3ce64d08707b029401c01f6a81ac0R61)
> 
> already deepcopies.

Given the assignment back later, I'd prefer not to shallow copy this.  I think it's just .DeepCopy, right?

pkg/operator/aws_platform_service_location/controller.go
```diff
@@ -38,12 +49,92 @@ func NewController(operatorClient operatorv1helpers.OperatorClient, infraClient
...
+		}
+	}
+	return allErrs.ToAggregate()
+}
+
+var schemeRE = regexp.MustCompile("^([^:]+)://")
+
+func validateServiceURL(uri string) error {
+	endpoint := uri
+	if !schemeRE.MatchString(endpoint) {
```
deads2k, line 123: This should really be a parseable URL or it shouldn't.  The sometimes API seems "nicer" but it ultimately makes for complication.  Let's do a straight `url.Parse` and then have a list of schemes we allow.  I don't think empty should be allowed and I would require https until we have a reason not to.

pkg/operator/aws_platform_service_location/controller.go
```diff
@@ -38,12 +49,92 @@ func NewController(operatorClient operatorv1helpers.OperatorClient, infraClient
...
+	}
+	if len(platformName) == 0 {
+		syncCtx.Recorder().Warningf("AWSPlatformServiceLocationController", "Falling back to deprecated status.platform because infrastructures.%s/cluster status.platformStatus.type is empty", configv1.GroupName)
+		platformName = currentInfra.Status.Platform
+	}
+
+	if platformName != configv1.AWSPlatformType {
+		return nil // nothing to do here.
+	}
+
```
deads2k, line 74: if the spec.platformspec.type != "" && spec.platformspec.type != platformName, it should error.

pkg/operator/aws_platform_service_location/controller.go
abhinavdahiya: `ServiceEndpoints` is a slice, so no DeepCopy.

i could DeepCopy the entire currentInfra again..
`currentInfra.DeepCopy().Spec.PlatformSpec.AWS.ServiceEndpoints` ?? @deads2k 

Can you provide a little background on why you think deepcopy here would be necessary?

deads2k: > `ServiceEndpoints` is a slice, so no DeepCopy.
> 
> i could DeepCopy the entire currentInfra again..
> `currentInfra.DeepCopy().Spec.PlatformSpec.AWS.ServiceEndpoints` ?? @deads2k
> 
> Can you provide a little background on why you think deepcopy here would be necessary?

mutating the input means that if you later use it or diff it somewhere, it won't be what you expect it to be.  Makes it hard for me to reason about it's state during this review.

PR #120, opened 2020-04-09
operator: add kube_cloud_config controller

abhinavdahiya:
The kube_cloud_config controller

- performs transformations on the user provided cloud config to create a new cloud config that
  stitches in the informatino from infrastructure.config.openshift.io object
- it creates and manages a stitched cloud config config-map and updates the status of the infra object
- it provides a way to have per platform transformers for stitching cloud config

There are 2 transformers for now,
- as-is
This copies all the contents of the source config map to target
This is used when there are no special transformers and the spec one needs to copied as-is

- aws
This transformer uses the service endpoints set in the infra.status.platformStatus.aws.serviceEndpoints to add
ServiceOverrides sections
If both user provided cloud-config and the infra have service endpoints, it returns an error

There is no Marshaller for gcfg used by the aws kube cloud-provided codebase, therefore we create ServiceOverrides sections using text/template and append those to the preexiting cloud config data.
It copies all the other keys in the config map as-is and only modifies the cloud config containing key.


go.mod
```diff
@@ -13,9 +13,15 @@ require (
...
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.4.0
 	golang.org/x/net v0.0.0-20200324143707-d3edc9973b7e // indirect
+	gopkg.in/gcfg.v1 v1.2.0
+	k8s.io/api v0.18.0
 	k8s.io/apimachinery v0.18.0
 	k8s.io/client-go v0.18.0-beta.2
 	k8s.io/component-base v0.18.0-beta.2
 	k8s.io/klog v1.0.0
+	k8s.io/legacy-cloud-providers v0.18.0-beta.2
```
abhinavdahiya, line 24: The reason i have decided to vendor in the module is because it allows us to keep bumping the API used for cloud-config otherwise we would be stuck carry a copy..

deads2k: > The reason i have decided to vendor in the module is because it allows us to keep bumping the API used for cloud-config otherwise we would be stuck carry a copy..

Just that one struct? We'd rather have the copy.  These are legacy for a reason and we want to be in control of our API surface.

pkg/operator/starter.go
```diff
@@ -21,15 +22,27 @@ import (
...
 	if err != nil {
 		return err
 	}
 	configInformers := configv1informers.NewSharedInformerFactory(configClient, 10*time.Minute)
-
+	kubeInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(kubeClient,
+		"",
+		operatorclient.GlobalUserSpecifiedConfigNamespace,
+		operatorclient.GlobalMachineSpecifiedConfigNamespace,
+		"kube-system",
```
deads2k, line 44: remove

pkg/operator/kube_cloud_config/aws.go
```diff
@@ -0,0 +1,100 @@
...
+
+	var region string
+	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.AWS != nil {
+		region = infra.Status.PlatformStatus.AWS.Region
+	}
+	if region == "" {
+		return nil, fmt.Errorf("Required region for AWS platform infrastructures.%s/cluster not found", configv1.GroupName)
+	}
+
+	output := input.DeepCopy()
```
deads2k, line 33: update namespace here to avoid overwrites.

pkg/operator/kube_cloud_config/aws_test.go
```diff
@@ -0,0 +1,353 @@
+package kube_cloud_config
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
```
deads2k, line 7: nit: no spaces.  gofmt doesn't auto-do it and we decided the separate groups didn't buy us anything.

pkg/operator/kube_cloud_config/controller.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/openshift/cluster-config-operator/pkg/operator/operatorclient"
+)
+
+const (
+	targetConfigName = "kube-cloud-config"
+	targetConfigKey  = "cloud.conf"
+)
+
+// KubeCloudConfigController is responsible for managing the kube-cloud-config used by various Kubernetes components
```
deads2k, line 29: tell me what this writes?

pkg/operator/kube_cloud_config/controller.go
```diff
@@ -0,0 +1,143 @@
...
+	if len(sourceCloudConfigMap) > 0 {
+		obj, err := c.configMapClient.ConfigMaps(operatorclient.GlobalUserSpecifiedConfigNamespace).Get(ctx, sourceCloudConfigMap, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		obj.DeepCopyInto(source)
+		source.ObjectMeta = metav1.ObjectMeta{}
+	}
+
+	t, ok := c.transformers[platformName]
```
deads2k, line 99: `cloudConfigTranformFn, ok` please

pkg/operator/kube_cloud_config/controller.go
```diff
@@ -0,0 +1,143 @@
...
+
+// KubeCloudConfigController is responsible for managing the kube-cloud-config used by various Kubernetes components
+// as source for configuration on clouds/platform.
+type KubeCloudConfigController struct {
+	infraClient     configv1client.InfrastructureInterface
+	infraLister     configv1listers.InfrastructureLister
+	configMapClient corev1client.ConfigMapsGetter
+
+	// transformers stores per platform tranformer
+	transformers map[configv1.PlatformType]transformer
```
deads2k, line 37: cloudConfigTransformers please

pkg/operator/kube_cloud_config/controller.go
```diff
@@ -0,0 +1,143 @@
...
+		t = asIsTransformer
+	}
+
+	target, err := t(source, sourceCloudConfigMapKey, currentInfra)
+	if err != nil {
+		return err
+	}
+
+	targetCloudConfigMap := targetConfigName
+	targetCloudConfigMapKey := sourceCloudConfigMapKey
```
deads2k, line 110: I prefer being prescriptive here.  I don't see a reason why we can't be.  If there is one, please add a comment.

pkg/operator/kube_cloud_config/controller.go
```diff
@@ -0,0 +1,143 @@
...
+	c := &KubeCloudConfigController{
+		infraClient:     infraClient.Infrastructures(),
+		infraLister:     infraLister,
+		configMapClient: configMapClient,
+		transformers: map[configv1.PlatformType]transformer{
+			configv1.AWSPlatformType: awsTransformer,
+		},
+	}
+	return factory.New().
+		WithInformers(
```
deads2k, line 54: need an informer for the source and destination configmap namespaces.

deads2k:
Need to add an OWNERS file with yourself as a reviewer and whoever else you'd like to review.  I think we'd like to leave approval powers top level for now, but may be comfortable relaxing it later.

pkg/operator/kube_cloud_config/controller.go
abhinavdahiya: hmm.. i'm not completely sure what you mean by the comment.

the key in the configmap that contains the cloud.conf is controlled by the sourceconfigmapkey. if there is no source, but we need a target config map we will use the _controlled controlled default key_ see https://github.com/openshift/cluster-config-operator/pull/120/files/650826a9518b5890af3b43ed27e1262ba53e2553..fe8c44193ee5245e7c735a94f81bf16099fa3559#diff-99d1c645e7848874c8b0c0ec171e00e1R85-R88

do you mean we should enforce that the key should always be `cloud.conf` ? that would have to be part of the infra api right??

go.mod
abhinavdahiya: done, copied the config type in `/kube_cloud_config/internal/aws`

pkg/operator/kube_cloud_config/aws.go
abhinavdahiya: > https://github.com/openshift/cluster-config-operator/pull/120/commits/626e8aae7dc1ccee9268753fd17744507e2d2603#diff-99d1c645e7848874c8b0c0ec171e00e1R27-R29

the transformer states that only the data and binaryData will be used by the consumer. the consumer has control over those fields.

pkg/operator/kube_cloud_config/controller.go
abhinavdahiya: updated the comment to provide more details.

go.mod
```diff
@@ -7,15 +7,21 @@ require (
...
+	k8s.io/api v0.18.0
 	k8s.io/apimachinery v0.18.0
-	k8s.io/client-go v0.18.0-beta.2
-	k8s.io/component-base v0.18.0-beta.2
+	k8s.io/client-go v0.18.0
+	k8s.io/component-base v0.18.0
 	k8s.io/klog v1.0.0
 )
+
+replace github.com/openshift/api => github.com/abhinavdahiya/api v0.0.0-20200413225449-37600d50200d
```
mfojtik, line 27: i would mark this PR as

/wip

until this is removed?

pkg/operator/kube_cloud_config/aws.go
deads2k: > > [626e8aa#diff-99d1c645e7848874c8b0c0ec171e00e1R27-R29](https://github.com/openshift/cluster-config-operator/commit/626e8aae7dc1ccee9268753fd17744507e2d2603#diff-99d1c645e7848874c8b0c0ec171e00e1R27-R29)
> 
> the transformer states that only the data and binaryData will be used by the consumer. the consumer has control over those fields.

belt and suspenders please

pkg/operator/kube_cloud_config/aws.go
```diff
@@ -0,0 +1,102 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-config-operator/pkg/operator/kube_cloud_config/internal/aws"
+	"gopkg.in/gcfg.v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// awsTransformer implements the cloudConfigTransformer. It uses the input ConfigMap and infra.status.platformStatus.AWS.ServiceEndpoints
+// to create a new config that include the ServiceOverrides sections. The transformer uses infra.status.platformStatus.region as the
+// signing service for all the ServiceOverrides.
+// It returns an error if the platform is AWSPlatformType.
```
deads2k, line 17: is not

pkg/operator/kube_cloud_config/aws.go
```diff
@@ -0,0 +1,102 @@
...
+func awsTransformer(input *corev1.ConfigMap, key string, infra *configv1.Infrastructure) (*corev1.ConfigMap, error) {
+	if !(infra.Status.PlatformStatus != nil &&
+		infra.Status.PlatformStatus.Type == configv1.AWSPlatformType) {
+		return nil, fmt.Errorf("invalid platform, expected to be AWS")
+	}
+	if !(infra.Status.PlatformStatus.AWS != nil && len(infra.Status.PlatformStatus.AWS.ServiceEndpoints) > 0) {
```
deads2k, line 23: demorgans please

pkg/operator/kube_cloud_config/aws.go
```diff
@@ -0,0 +1,102 @@
...
+		return asIsTransformer(input, key, infra) // no transformation required
+	}
+
+	var region string
+	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.AWS != nil {
+		region = infra.Status.PlatformStatus.AWS.Region
+	}
+	if region == "" {
+		return nil, fmt.Errorf("Required region for AWS platform infrastructures.%s/cluster not found", configv1.GroupName)
```
deads2k, line 32: make this into a field.Required or some such

pkg/operator/kube_cloud_config/aws.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if len(inCfgRaw.String()) > 0 {
+		var cfg aws.CloudConfig
+		err := gcfg.ReadInto(&cfg, bytes.NewBufferString(inCfgRaw.String()))
+		if err != nil {
+			return nil, fmt.Errorf("failed to read the cloud.conf: %w", err)
+		}
+
+		if len(cfg.ServiceOverride) > 0 {
```
deads2k, line 51: What is the downside to checking to see if the values conflict?  A cluster-admin may well have an existing config file that already specifies all the same information, right?

abhinavdahiya: because I think it's safer for us to ensure that there is only one way to configure a thing. Allowing user to service override and also in spec means there is duplication of the same information.

Also since there is no marshaller for gcfg _the format used for this configuration file upstream_, there is no sane way to update user specified to match the correct standards in case there is pre-existing.

PR #121, opened 2020-04-14
operator: move config metrics from kube-apiserver operator

mfojtik:
Move https://github.com/openshift/cluster-kube-apiserver-operator/tree/master/pkg/operator/configmetrics to this operator as this is the logical place to have them. 

I'm interested how the prometheus will handle this transition, when for short time both metrics will exists.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/openshift/cluster-config-operator
 go 1.13
 
 require (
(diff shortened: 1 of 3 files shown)


deads2k:
/hold

try opening that here first: https://github.com/openshift/openshift-state-metrics

PR #123, opened 2020-04-15
version: fix build_info metric reporting

mfojtik:


pkg/version/version.go
```diff
@@ -19,29 +17,25 @@ var (
...
+		GoVersion:    goVersion,
+		Compiler:     compiler,
+		Platform:     platform,
 	}
 }
-
-func init() {
-	buildInfo := metrics.NewGaugeVec(
-		&metrics.GaugeOpts{
-			Name: "openshift_cluster_config_operator_build_info",
```
sttts, line 39: where is this constant gone? Are we sure we don't change the gauge name?

PR #126, opened 2020-04-20
Bug 1825944: kube_cloud_config: enforce the target key in config map to cloud.conf

abhinavdahiya:
The infrastructure.config.openshift.io/v1 requires that the key for cloud config in the generated config map be `cloud.conf`.
Previously the controller was just copying and not enforcing the target key.

no both the as-is and aws transformers enforce this, and also the tests were updated to check that.


abhinavdahiya:
https://github.com/openshift/release/pull/8437 should help run azure, gcp clusters for testing.

PR #127, opened 2020-04-20
Bug 1814332: add migration_aws_status controller

abhinavdahiya:
This controller is responsible for upgrading existing `infrastructure.config.openshift.io/v1` `cluster` object
for AWS to include the `AWSPlatformStatus` based on the InstallConfig stored in the ConfigMap `kube-system/cluster-config-v1`.
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1814332

/cc @deads2k @smarterclayton


pkg/operator/migration_aws_status/controller.go
```diff
@@ -0,0 +1,131 @@
...
+		syncCtx.Recorder().Warningf("MigrationAWSStatusController", "Required infrastructures.%s/cluster not found", configv1.GroupName)
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	currentInfra := obji.DeepCopy()
+
+	if currentInfra.Status.Platform != configv1.AWSPlatformType || // not aws
```
wking, line 77: [`Status.Platform` is deprecated][1].  I'd rather have something like:

```go
var platformType configv1.PlatformType
if currentInfra.Status.PlatformStatus != nil && len(currentInfra.Status.PlatformStatus.Type) >  0 {
  platformType = currentInfra.Status.PlatformStatus.Type
} else {
  platformType = currentInfra.Status.Platform
}

if platformType != configv1.AWSPlatformType || // not AWS
    ... {
  ...
}
```

or some such.

[1]: https://github.com/openshift/api/blob/0f159fee64dbf711d40dac3fa2ec8b563a2aaca8/config/v1/types_infrastructure.go#L54-L57

pkg/operator/migration_aws_status/controller.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	config := installConfg{}
+	if err := yaml.Unmarshal([]byte(configRaw), &config); err != nil {
+		return installConfg{}, fmt.Errorf("unable to parse install-config.yaml: %s", err)
+	}
+	return config, nil
+}
+
+type installConfg struct {
```
wking, line 125: `installConfg` -> `installConfig`?

pkg/operator/migration_aws_status/controller.go
abhinavdahiya: deprecated fields don't mean they will not be set... to keep backward compatibility that field will always be set. the clients should start moving to the new fields.

And since this is the migrator that is moving clusters from deprecated to new state, and therefore not using the .status.platformType feels weird.. i don't think your if-else guard makes sense here.

smarterclayton: +100 here - deprecated means "you should start using something new" not "the old one changes or goes away".

We DO NOT remove fields ever. :)

pkg/operator/migration_aws_status/controller.go
```diff
@@ -0,0 +1,132 @@
...
+	currentInfra := obji.DeepCopy()
+
+	if currentInfra.Status.Platform != configv1.AWSPlatformType || // not aws
+		(currentInfra.Status.PlatformStatus != nil && currentInfra.Status.PlatformStatus.Type != configv1.AWSPlatformType) || // not aws
+		(currentInfra.Status.PlatformStatus != nil && currentInfra.Status.PlatformStatus.Type == configv1.AWSPlatformType &&
+			currentInfra.Status.PlatformStatus.AWS != nil && len(currentInfra.Status.PlatformStatus.AWS.Region) > 0) { // aws, but the region is already set. so no need to migrate
+		return nil // no action
+	}
+
+	cc, err := loadClusterConfig(ctx, c.configMapClient)
```
deads2k, line 85: if the cluster config isn't present, short-circuit please.  Our end goal should be removal.

pkg/operator/migration_aws_status/controller.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	currentInfra := obji.DeepCopy()
+
+	if currentInfra.Status.Platform != configv1.AWSPlatformType || // not aws
+		(currentInfra.Status.PlatformStatus != nil && currentInfra.Status.PlatformStatus.Type != configv1.AWSPlatformType) || // not aws
```
deads2k, line 79: wow pointers suck.  Why is this one a pointer again?  Someone didn't like optional fields with empty strings having meaning of empty?

pkg/operator/migration_aws_status/controller.go
deads2k: I see trevor's comment more as pointing out that if the authoritative field is .status.platformstatus.type (omg, why pointers??), we should trigger on that field and stomp the deprecated one to whatever is correct.

I'd like to have a goal of transitioning this field away from a pointer.  We can do that by requiring it in every cluster in 4.5.  This code highlights what's wrong with pointers.  Dear lord.

deads2k:
the idea is sound, I'll leave lgtm to someone of your choosing

/approve

pkg/operator/migration_aws_status/controller.go
abhinavdahiya: I understand @wking comment and i have already added similar conditonal to other controllers like kube_cloud_config, aws_service_endpoints.

I was trying to say that, for this specific controller the source of truth needs to be the deprecated field because it is trying to migrate the clusters away so that use of deprecated can be gone later on.

after this migration is applied to 4.5, we can make the move from pointer. so 4.6

wking: > I was trying to say that, for this specific controller the source of truth needs to be the deprecated field...

I would be very surprised if existing 4.5 consumers are checking the deprecated field first and then falling back to the new field if the deprecated field is unset.

wking: registry operator [only falls back to install-config if `PlatformStatus` is `nil`][1].

[1]: https://github.com/openshift/cluster-image-registry-operator/blob/24d4772d517571edccbd219c43f34954435c0c3b/pkg/storage/util/util.go#L76-L85

abhinavdahiya: @wking the registry operator is going the right thing. as it is supposed to used the new field and only when unset use the deprecated one.

but doesn't mean this migration needs to do that.

abhinavdahiya: > I would be very surprised if existing 4.5 consumers are checking the deprecated field first and then falling back to the new field if the deprecated field is unset.

again echoing the point from https://github.com/openshift/cluster-config-operator/pull/127#discussion_r412305448 https://github.com/openshift/cluster-config-operator/pull/127#discussion_r411780867

this deprecated field cannot go away unless we bump api version , and will always continue to be set for the life of the api. that how k8s backward compatibilty works.

pkg/operator/migration_aws_status/controller.go
abhinavdahiya: for this migration to run the cluster-config is required. so i don't think the short-circuit makes sense.

if the cluster-config is gone and this migration still needs to run, that's an error.
this migration controller is going to go away before we think about dropping the cluster-config

pkg/operator/migration_aws_status/controller.go
```diff
@@ -0,0 +1,132 @@
...
+		return fmt.Errorf("empty region set in cluster-config-v1")
+	}
+
+	currentInfra.Status.PlatformStatus = &configv1.PlatformStatus{
+		Type: configv1.AWSPlatformType,
+		AWS:  &configv1.AWSPlatformStatus{Region: cc.Platform.AWS.Region},
+	}
+	_, err = c.infraClient.UpdateStatus(ctx, currentInfra, metav1.UpdateOptions{})
+	if err != nil {
+		syncCtx.Recorder().Warningf("MigrationAWSStatusController", "Unable to updated the infrastructure status")
```
patrickdillon, line 103: `updated` -> `update`

pkg/operator/migration_aws_status/controller.go
wking: > but doesn't mean this migration needs to do that.

If this PR prefers the deprecated property while the registry prefers the new property, then in the (unlikely) event of divergence, this migration is going to move the registry to a new region (not actually sure what the registry operator does if the region changes under its feet).  I think we want to consistently have folks prefer the new property and fall back to the deprecated property, and we also want this logic to update the deprecated property if it diverges, so old tools that only consume the deprecated property don't see stale data.

PR #130, opened 2020-05-15
Bug 1784151: set additionalPrinterColumns for SCC

soltysh:
/assign @sttts 

This is fake bump to pick https://github.com/openshift/api/pull/648
/hold


soltysh:
/hold cancel
now that the bump is good.
@sttts ptal

PR #131, opened 2020-05-19
Bug 1825194: oc explain still shows "The default is auth.openshift.io"

p0lyn0mial:


p0lyn0mial:
I've just verified it actually works on a cluster created by the cluster-bot (`launch openshift/cluster-config-operator#131`)

```
 oc explain authentication.spec --api-version config.openshift.io/v1
KIND:     Authentication
VERSION:  config.openshift.io/v1

RESOURCE: spec <Object>

DESCRIPTION:
     spec holds user settable values for configuration

FIELDS:
   oauthMetadata        <Object>
     oauthMetadata contains the discovery endpoint data for OAuth 2.0
     Authorization Server Metadata for an external OAuth server. This discovery
     document can be viewed from its served location: oc get --raw
     '/.well-known/oauth-authorization-server' For further details, see the IETF
     Draft: https://tools.ietf.org/html/draft-ietf-oauth-discovery-04#section-2
     If oauthMetadata.name is non-empty, this value has precedence over any
     metadata reference stored in status. The key "oauthMetadata" is used to
     locate the data. If specified and the config map or expected key is not
     found, no metadata is served. If the specified metadata is not valid, no
     metadata is served. The namespace for this config map is openshift-config.

   serviceAccountIssuer <string>
     serviceAccountIssuer is the identifier of the bound service account token
     issuer. The default is https://kubernetes.default.svc

   type <string>
     type identifies the cluster managed, user facing authentication mode in
     use. Specifically, it manages the component that responds to login
     attempts. The default is IntegratedOAuth.

   webhookTokenAuthenticators   <[]Object>
     webhookTokenAuthenticators configures remote token reviewers. These remote
     authentication webhooks can be used to verify bearer tokens via the
     tokenreviews.authentication.k8s.io REST API. This is required to honor
     bearer tokens that are provisioned by an external authentication service.
     The namespace for these secrets is openshift-config
```

mfojtik:
patch manager note: can you please pick this up for next z-stream?

PR #133, opened 2020-05-20
kube_cloud_config: support for other Azure cloud environments

staebler:
Populate the `cloud` field in kube cloud.conf stored in the kube-cloud-config ConfigMap in the openshift-config-managed namespace. This field informs in which Azure cloud environment the cluster is installed and consequently which Azure API endpoint should be used when  communicating via the Azure SDK.

The value populated comes from the `.status.platformStatus.azure.cloudName` field of the infrastructure.config.openshift.io resource. If the field is empty or missing, then the default "AZUREPUBLICCLOUD" value is set for the `cloud` field in the kube cloud config. If the field generated from the infrastructure resource conflicts with the field in the user-provided kube
cloud config, then the controller will error on syncing the infrastructure resource.

See openshift/enhancements#321

https://issues.redhat.com/browse/CORS-1444

### DIFF
--- a/pkg/operator/kube_cloud_config/azure.go
+++ b/pkg/operator/kube_cloud_config/azure.go
@@ -1,0 +1,119 @@
+package kube_cloud_config
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"sigs.k8s.io/yaml"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	"github.com/openshift/cluster-config-operator/pkg/operator/operatorclient"
+)
+
+const (
+	azureCloudFieldName = "cloud"
+)
+
+var (
+	validAzureCloudNames = map[configv1.AzureCloudEnvironment]bool{
+		configv1.AzurePublicCloud:       true,
+		configv1.AzureUSGovernmentCloud: true,
+		configv1.AzureChinaCloud:        true,
+		configv1.AzureGermanCloud:       true,
+	}
+
+	validAzureCloudNameValues = func() []string {
+		v := make([]string, 0, len(validAzureCloudNames))
+		for n := range validAzureCloudNames {
+			v = append(v, string(n))
+		}
+		return v
+	}()
+)
+
+// azureTransformer implements the cloudConfigTransformer. It uses the input ConfigMap and infra.status.platformStatus.azure.cloudName
+// to create a new config that has the cloud field set.
+// It returns an error if the platform is not AzurePlatformType.
+func azureTransformer(input *corev1.ConfigMap, key string, infra *configv1.Infrastructure) (*corev1.ConfigMap, error) {
+	if !(infra.Status.PlatformStatus != nil &&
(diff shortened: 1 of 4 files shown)


staebler:
/hold

This is currently vendoring an openshift/api fork until https://github.com/openshift/api/pull/650 merges.

staebler:
346a2aa5...7f5aeffb
* Changed to not force the `cloud` field in the kube cloud config to uppercase.

staebler:
7f5aeffb...9656e5d8
* Bumped to the latest master version of openshift/api to include the changes from openshift/api#650.
* Rebased onto latest master

/hold cancel

staebler:
/hold

The machine config is not getting generated. The masters are stuck using 99-master-generated-registries.

staebler:
> The machine config is not getting generated. The masters are stuck using 99-master-generated-registries.

The cluster-config-operator is writing out a different cloud.conf, which is causing the machine-config-operator to render a different machineconfig. The machine-config daemon running in bootstrap node is looking for the first hash, before the cluster-config-operator changes.

staebler:
9656e5d8...1a7e4e00
* Changed the Azure transformer to encode the cloud.conf using the same method that the installer does, so that the cloud.conf is not changed during bootstrap. The transformer had been encoding in alphabetical order and without tabs and newlines. The installer encodes in field order with tabs and newlines.

staebler:
1a7e4e00...7299eac1
* Moved last changed to the correct commit.

staebler:
The azure tests look to be failing to create the worker nodes. I assume this is due to the bug where the validation is rejecting changes to the machine when `providerSpec.userDataSecret.namespace` is empty.

This may have been the underlying cause of the masters not being able to transition to the new state after the cloud.conf was changed. So my recent changes to this PR may not have been necessary. I think it is better that the cluster-config-operator does not change the cloud.conf if it does not have to. But this does require copying type from https://github.com/kubernetes/kubernetes/blob/v1.13.5/pkg/cloudprovider/providers/azure. I had been hesitant to do that since it introduces something else that may become out-of-sync and is not easy to discover. I could change it to use the incoming cloud.conf unmodified when the cloud already matches and fall back to alphabetical ordering when a change is necessary?

abhinavdahiya:
> https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-config-operator/133/pull-ci-openshift-cluster-config-operator-master-e2e-azure/1275107928658939904

shows the installs are completing..

/hold cancel

go.mod
```diff
@@ -15,8 +15,8 @@ require (
 	github.com/stretchr/testify v1.4.0
 	gopkg.in/gcfg.v1 v1.2.0
 	gopkg.in/warnings.v0 v0.1.1 // indirect
-	k8s.io/api v0.18.2
-	k8s.io/apimachinery v0.18.2
+	k8s.io/api v0.18.3
+	k8s.io/apimachinery v0.18.3
```
sttts, line 19: why version skew?

sttts:
Needs rebase.

See version skew nit.

Otherwise, sgtm.

staebler:
> Needs rebase.
> 
> See version skew nit.
> 
> Otherwise, sgtm.

The version of openshift/api has already been bumped in master, so the commit bumping the version in this PR has been removed.

PR #135, opened 2020-06-16
Fix 4.4 build failure

vfreex:
We got the following  error from the latest 4.4 ose-cluster-config-operator build:
```
2020-06-15 14:42:23,127 - atomic_reactor.plugins.imagebuilder - INFO - --> COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/config/v1/*_config-operator_*-.yaml /usr/share/bootkube/manifests/manifests
2020-06-15 14:42:23,884 - atomic_reactor.plugins.imagebuilder - INFO - file does not exist
```
(extracted from http://download.eng.bos.redhat.com/brewroot/work/tasks/9876/29399876/x86_64.log)
Seems to me that `https://github.com/openshift/cluster-config-operator/blob/release-4.4/Dockerfile.rhel7#L10` has an extra `-` before .yaml . Not sure if my guessing is true or anything is missing. Could you guys have a look?

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -7,7 +7,7 @@ RUN GODEBUG=tls13=1 go build -ldflags "-X $GO_PACKAGE/pkg/version.versionFromGit
 FROM registry.svc.ci.openshift.org/ocp/4.2:base
 RUN mkdir -p /usr/share/bootkube/manifests/manifests
 RUN mkdir -p /usr/share/bootkube/manifests/bootstrap-manifests
-COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/config/v1/*_config-operator_*-.yaml /usr/share/bootkube/manifests/manifests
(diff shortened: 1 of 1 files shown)


tnozicka:
Thanks for catching this!

I've send PRs to fix our CI configuration to use the same Dockerfile as ART https://github.com/openshift/release/pull/9679 to prevent diverging like this.

Your PR needs to start with master branch though and then be backported. It's also gonna be required to have a BZ attached for master and then clones of that BZ for 4.5 and 4.4 always waiting for the previous one to be verified by QA first before a merge.

tnozicka:
Since the release PR merged to use the .rhel7 dockerfile to use the same one as ART so we are in a bit of a hurry to fix this, so I've created the chain of PRs also including your fix.

https://github.com/openshift/cluster-config-operator/pull/136
https://github.com/openshift/cluster-config-operator/pull/137
https://github.com/openshift/cluster-config-operator/pull/138

thanks again.
/close
(in favour of those PRs)

PR #136, opened 2020-06-16
Bug 1847455: Fix Dockerfiles

tnozicka:
With https://github.com/openshift/release/pull/9679 CI now uses the correct dockerfile so we can remove the stale one. Also taking the typo fix from https://github.com/openshift/cluster-config-operator/pull/135

/cc @sttts


Dockerfile.rhel7
```diff
@@ -7,7 +7,7 @@ RUN make build --warn-undefined-variables
 FROM registry.svc.ci.openshift.org/ocp/4.2:base
 RUN mkdir -p /usr/share/bootkube/manifests/manifests
 RUN mkdir -p /usr/share/bootkube/manifests/bootstrap-manifests
-COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/config/v1/*_config-operator_*-.yaml /usr/share/bootkube/manifests/manifests
+COPY --from=builder /go/src/github.com/openshift/cluster-config-operator/vendor/github.com/openshift/api/config/v1/*_config-operator_*.yaml /usr/share/bootkube/manifests/manifests
```
sttts, line 10: is there no variant which fails if there are no files?

tnozicka: some builder fails, not ours. this should have been caught by CI except it was using the wrong dockerfile so this one wasn't tested

PR #140, opened 2020-06-19
render: generate cloud config when infrastructure and configmap are provided

jhixson74:
This relies on https://github.com/openshift/cluster-config-operator/pull/133 and is included for now. When it can be shown that it works, I can remove those commits from this PR. This is very much a WIP, so please point out anywhere that I am going about things completely wrong ;-)

https://issues.redhat.com/browse/CORS-1458

### DIFF
--- a/pkg/cmd/render/render.go
+++ b/pkg/cmd/render/render.go
@@ -1,6 +1,7 @@
 package render
 
 import (
+	"fmt"
 	"io/ioutil"
 
 	"github.com/spf13/cobra"
@@ -8,6 +9,7 @@ import (
 	"k8s.io/klog"
 
 	configv1 "github.com/openshift/api/config/v1"
+	kubecloudconfig "github.com/openshift/cluster-config-operator/pkg/operator/kube_cloud_config"
 	genericrender "github.com/openshift/library-go/pkg/operator/render"
 	genericrenderoptions "github.com/openshift/library-go/pkg/operator/render/options"
 )
@@ -18,6 +20,10 @@ type renderOpts struct {
 	generic  genericrenderoptions.GenericOptions
 
 	clusterConfigFile string
+
+	clusterInfrastructureInputFile string
+	cloudProviderConfigInputFile   string
+	cloudProviderConfigOutputFile  string
 }
 
 // NewRenderCommand creates a render command.
@@ -52,6 +58,16 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	r.generic.AddFlags(fs, configv1.GroupVersion.WithKind("Config"))
 
 	fs.StringVar(&r.clusterConfigFile, "cluster-config-file", r.clusterConfigFile, "Openshift Cluster API Config file.")
+
+	// This is the file containing the infrastructure object
+	fs.StringVar(&r.clusterInfrastructureInputFile, "cluster-infrastructure-input-file", r.clusterInfrastructureInputFile, "Input path for the cluster infrastructure file.")
+
+	// This is the file containing the configmap for the kube cloud config provided by the user
+	fs.StringVar(&r.cloudProviderConfigInputFile, "cloud-provider-config-input-file", r.cloudProviderConfigInputFile, "Input path for the cloud provider config file.")
+
+	// This is the generated kube cloud config
+	fs.StringVar(&r.cloudProviderConfigOutputFile, "cloud-provider-config-output-file", r.cloudProviderConfigOutputFile, "Output path for the generated cloud provider config file.")
+
 }
 
 // Validate verifies the inputs.
@@ -61,6 +77,18 @@ func (r *renderOpts) Validate() error {
 	}
 	if err := r.generic.Validate(); err != nil {
 		return err
+	}
+
+	// Validate all files are specified when specifying infrastructure and configmap files
+	if infra, provider := len(r.clusterInfrastructureInputFile) != 0, len(r.cloudProviderConfigOutputFile) != 0; infra || provider {
+		if !(infra && provider) {
+			return fmt.Errorf("clulster-infrastructure-file and cloud-provider-config-output-file must be specified.")
+		}
+		if infra {
+			if err := kubecloudconfig.ValidateFile(r.clusterInfrastructureInputFile); err != nil {
+				return err
+			}
+		}
 	}
 
 	return nil
@@ -85,6 +113,7 @@ type TemplateData struct {
 // Run contains the logic of the render command.
 func (r *renderOpts) Run() error {
 	renderConfig := TemplateData{}
+
 	if len(r.clusterConfigFile) > 0 {
 		_, err := ioutil.ReadFile(r.clusterConfigFile)
 		if err != nil {
@@ -92,21 +121,33 @@ func (r *renderOpts) Run() error {
 		}
 		// TODO I'm thinking we parse this into a map and reference it that way
 	}
+
 	if err := r.manifest.ApplyTo(&renderConfig.ManifestConfig); err != nil {
 		return err
 	}
 	if err := r.generic.ApplyTo(
 		&renderConfig.FileConfig,
-
-		// TODO I don't think we have these
 		genericrenderoptions.Template{},
 		genericrenderoptions.Template{},
-
 		&renderConfig,
 		nil,
 	); err != nil {
 		return err
 	}
 
-	return genericrender.WriteFiles(&r.generic, &renderConfig.FileConfig, renderConfig)
+	if err := genericrender.WriteFiles(&r.generic, &renderConfig.FileConfig, renderConfig); err != nil {
+		return err
+	}
+
+	if len(r.clusterInfrastructureInputFile) > 0 && len(r.cloudProviderConfigOutputFile) > 0 {
+		targetCloudConfigMapData, err := kubecloudconfig.BootstrapTransform(r.clusterInfrastructureInputFile, r.cloudProviderConfigInputFile)
+		if err != nil {
+			return err
+		}
+		if err := ioutil.WriteFile(r.cloudProviderConfigOutputFile, targetCloudConfigMapData, 0644); err != nil {
+			return fmt.Errorf("failed to write merged config to %q: %v", r.cloudProviderConfigOutputFile, err)
+		}
+	}
+
+	return nil
 }
--- a/pkg/operator/kube_cloud_config/bootstrap.go
+++ b/pkg/operator/kube_cloud_config/bootstrap.go
@@ -1,0 +1,89 @@
+package kubecloudconfig
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/yaml"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorclient "github.com/openshift/cluster-config-operator/pkg/operator/operatorclient"
+)
+
+// ValidateFile verifies a file exists, has content, and is a regular file
+func ValidateFile(path string) error {
+	st, err := os.Stat(path)
+	if err != nil {
+		return fmt.Errorf("failed to stat %s: %w", path, err)
+	}
+	if !st.Mode().IsRegular() {
+		return fmt.Errorf("%s is not a regular file", path)
+	}
+	if st.Size() <= 0 {
+		return fmt.Errorf("%s is empty", path)
+	}
+	return nil
+}
+
+// BootstrapTransform implements the cloudConfigTransformer during bootstrapping.
+// It uses the input ConfigMap and Infrastructure provided by files on the bootstrap
+// host to create a new config that has the cloud field set.
+func BootstrapTransform(infrastructureFile string, cloudProviderFile string) ([]byte, error) {
+
+	// Read, parse, and save the infrastructure object
+	var clusterInfrastructure configv1.Infrastructure
+	fileData, err := ioutil.ReadFile(infrastructureFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read infrastructure file: %w", err)
+	}
+	err = yaml.Unmarshal(fileData, &clusterInfrastructure)
+	if err != nil {
+		return nil, fmt.Errorf("failed unmarshal infrastructure: %w", err)
+	}
+
+	// Read, parse, and save the user provided cloud configmap
+	var cloudProviderConfigInput corev1.ConfigMap
+	if len(cloudProviderFile) > 0 {
+		fileData, err = ioutil.ReadFile(cloudProviderFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read cloud provider file: %w", err)
+		}
+		err = yaml.Unmarshal(fileData, &cloudProviderConfigInput)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal cloud provider: %w", err)
+		}
+	}
+
+	// Determine the platform type
+	var platformName configv1.PlatformType
+	if pstatus := clusterInfrastructure.Status.PlatformStatus; pstatus != nil {
+		platformName = pstatus.Type
+	}
+	if len(platformName) == 0 {
+		platformName = clusterInfrastructure.Status.Platform
+	}
+
+	// Determine the platform specific transformer method to use
+	cloudConfigTransformers := cloudConfigTransformers()
+	cloudConfigTransformerFn, ok := cloudConfigTransformers[platformName]
+	if !ok {
+		cloudConfigTransformerFn = asIsTransformer
+	}
+	target, err := cloudConfigTransformerFn(&cloudProviderConfigInput, clusterInfrastructure.Spec.CloudConfig.Key, &clusterInfrastructure)
+	if err != nil {
+		return nil, fmt.Errorf("failed to transform cloud config: %w", err)
+	}
+
+	target.Name = targetConfigName
+	target.Namespace = operatorclient.GlobalMachineSpecifiedConfigNamespace
+	/* ApplyConfigMap() */
+
+	targetCloudConfigMapData, err := yaml.Marshal(target)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marhsal cloud config: %w", err)
+	}
+
+	return targetCloudConfigMapData, nil
+}
--- a/pkg/operator/kube_cloud_config/controller.go
+++ b/pkg/operator/kube_cloud_config/controller.go
@@ -1,4 +1,4 @@
-package kube_cloud_config
+package kubecloudconfig
 
 import (
 	"context"
@@ -49,13 +49,10 @@ func NewController(operatorClient operatorv1helpers.OperatorClient,
 	openshiftConfigConfigMapInformer cache.SharedIndexInformer, openshiftConfigManagedConfigMapInformer cache.SharedIndexInformer,
 	recorder events.Recorder) factory.Controller {
 	c := &KubeCloudConfigController{
-		infraClient:     infraClient.Infrastructures(),
-		infraLister:     infraLister,
-		configMapClient: configMapClient,
-		cloudConfigTransformers: map[configv1.PlatformType]cloudConfigTransformer{
-			configv1.AWSPlatformType:   awsTransformer,
-			configv1.AzurePlatformType: azureTransformer,
-		},
+		infraClient:             infraClient.Infrastructures(),
+		infraLister:             infraLister,
+		configMapClient:         configMapClient,
+		cloudConfigTransformers: cloudConfigTransformers(),
 	}
 	return factory.New().
 		WithInformers(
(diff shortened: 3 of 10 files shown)


pkg/cmd/render/render.go
```diff
@@ -63,6 +84,41 @@ func (r *renderOpts) Validate() error {
...
 
+	clusterInfrastructureFileLen := len(r.clusterInfrastructureFile)
+	cloudProviderConfigInputFileLen := len(r.cloudProviderConfigInputFile)
+	cloudProviderConfigOutputFileLine := len(r.cloudProviderConfigOutputFile)
+
+	// Validate all files are specified when specifying infrastructure and configmap files
+	if clusterInfrastructureFileLen != 0 || cloudProviderConfigInputFileLen != 0 || cloudProviderConfigOutputFileLine != 0 {
+		if !(clusterInfrastructureFileLen != 0 && cloudProviderConfigInputFileLen != 0 && cloudProviderConfigOutputFileLine != 0) {
+			return fmt.Errorf("clulster-infrastructure-file, cloud-provider-config-input-file, and cloud-provider-config-output-file must all be specified.")
+		}
```
abhinavdahiya, line 95: there is no requirement for input cloud provider file. see AWS, when using service endpoints there is no input configmap but there is a generated one with all the service overrides.

I think we need to take into account that.

pkg/cmd/render/render.go
```diff
@@ -80,28 +136,98 @@ func (r *renderOpts) Complete() error {
...
+		targetCloudConfigMap := kubecloudconfig.TargetConfigName()
+
+		target.Name = kubecloudconfig.TargetConfigName()
+		target.Namespace = operatorclient.GlobalMachineSpecifiedConfigNamespace
+		/* ApplyConfigMap() */
+
+		targetCloudConfigMapData, err := yaml.Marshal(target)
+		if err != nil {
+			return err
+		}
```
abhinavdahiya, line 216: all this should be part of the kube_cloud_config package.

we can keep the load inputs in here, but how/when to generate the new cloud config file should be controlled by kube_cloud_config package. the sync also needs something like that so maybe we can re-use one implementation...

go.mod
```diff
@@ -15,10 +15,12 @@ require (
...
-	k8s.io/apimachinery v0.18.2
+	k8s.io/api v0.18.3
+	k8s.io/apimachinery v0.18.3
 	k8s.io/client-go v0.18.2
 	k8s.io/component-base v0.18.2
 	k8s.io/klog v1.0.0
 	sigs.k8s.io/yaml v1.2.0
 )
+
+replace github.com/openshift/api => github.com/staebler/api v0.0.0-20200520010057-f4bb75a9024c
```
sttts, line 26: this will be replace I assume.

pkg/cmd/render/render.go
```diff
@@ -52,6 +63,16 @@ func (r *renderOpts) AddFlags(fs *pflag.FlagSet) {
 	r.generic.AddFlags(fs, configv1.GroupVersion.WithKind("Config"))
 
 	fs.StringVar(&r.clusterConfigFile, "cluster-config-file", r.clusterConfigFile, "Openshift Cluster API Config file.")
+
+	// This is the file containing the infrastructure object
+	fs.StringVar(&r.clusterInfrastructureFile, "cluster-infrastructure-file", r.clusterInfrastructureFile, "Input path for the cluster infrastructure file.")
```
sttts, line 68: input or output file?  Either we should have that postfix everywhere or nowhere.

pkg/cmd/render/render.go
```diff
@@ -63,6 +84,41 @@ func (r *renderOpts) Validate() error {
 		return err
 	}
 
+	clusterInfrastructureFileLen := len(r.clusterInfrastructureFile)
+	cloudProviderConfigInputFileLen := len(r.cloudProviderConfigInputFile)
+	cloudProviderConfigOutputFileLine := len(r.cloudProviderConfigOutputFile)
```
sttts, line 89: I don't think these make it any easier to read. Just inline the use of len and drop the vars.

pkg/cmd/render/render.go
```diff
@@ -63,6 +84,41 @@ func (r *renderOpts) Validate() error {
...
+
+		// Validate infrastructure file exists, has content, and is a regular file
+		ci, err := os.Stat(r.clusterInfrastructureFile)
+		if err != nil {
+			return err
+		}
+		if ci.Size() <= 0 {
+			return fmt.Errorf("cluster-infrastructure-file cannot be empty.")
+		}
+		if !ci.Mode().IsRegular() {
```
sttts, line 105: would have expected opposite order: first mode, then size

pkg/cmd/render/render.go
```diff
@@ -80,28 +136,98 @@ func (r *renderOpts) Complete() error {
...
 			return err
 		}
 		// TODO I'm thinking we parse this into a map and reference it that way
 	}
+
+	if len(r.clusterInfrastructureFile) > 0 && len(r.cloudProviderConfigInputFile) > 0 && len(r.cloudProviderConfigOutputFile) > 0 {
+
+		// Read, parse, and save the infrastructure object
+		var clusterInfrastructure configv1.Infrastructure
+		clusterInfrastructureFileData, err := ioutil.ReadFile(r.clusterInfrastructureFile)
```
sttts, line 161: use short variable names. This looks like Java code. `bs` for bytes is commonly used.

pkg/cmd/render/render.go
```diff
@@ -80,28 +136,98 @@ func (r *renderOpts) Complete() error {
...
+		target.Name = kubecloudconfig.TargetConfigName()
+		target.Namespace = operatorclient.GlobalMachineSpecifiedConfigNamespace
+		/* ApplyConfigMap() */
+
+		targetCloudConfigMapData, err := yaml.Marshal(target)
+		if err != nil {
+			return err
+		}
+
+		if err := ioutil.WriteFile(r.cloudProviderConfigOutputFile, targetCloudConfigMapData, 0644); err != nil {
```
sttts, line 218: little odd to have this write up here. Would expect it at the end of this render func.

go.mod
jhixson74: Updated.

pkg/cmd/render/render.go
jhixson74: I've changed the names to reflect input/output.

pkg/cmd/render/render.go
jhixson74: Removed.

pkg/cmd/render/render.go
jhixson74: I've changed the order.

pkg/cmd/render/render.go
jhixson74: I've shortened the names. I am by no means a Java programmer, was just following conventions used in much of the golang code I've reviewed ;-)

pkg/cmd/render/render.go
jhixson74: I've moved it to the bottom of the function.

pkg/cmd/render/render.go
jhixson74: I've moved the code to kube_cloud_config. I still need to look at sync() and perhaps generalize the transformer functions also.

pkg/cmd/render/render.go
jhixson74: Still to-do.

pkg/cmd/render/render.go
```diff
@@ -85,28 +113,44 @@ type TemplateData struct {
...
 	}
 
-	return genericrender.WriteFiles(&r.generic, &renderConfig.FileConfig, renderConfig)
+	if err := genericrender.WriteFiles(&r.generic, &renderConfig.FileConfig, renderConfig); err != nil {
+		return err
+	}
+
+	//if len(r.clusterInfrastructureInputFile) > 0 && (r.cloudProviderConfigInputFile) > 0 && len(r.cloudProviderConfigOutputFile) > 0 {
+
+	if len(r.clusterInfrastructureInputFile) > 0 && len(r.cloudProviderConfigOutputFile) > 0 {
```
abhinavdahiya, line 145: don't we already require that the input output file names be provided https://github.com/openshift/cluster-config-operator/pull/140/files#diff-5e45df1052a09901f6732e8a838541ebR85? so is this check required?

pkg/operator/kube_cloud_config/transform.go
```diff
@@ -0,0 +1,99 @@
...
+
+func BootstrapTransformer(infrastructureFile string, cloudProviderFile string) ([]byte, error) {
+
+	// Read, parse, and save the infrastructure object
+	var clusterInfrastructure configv1.Infrastructure
+	fileData, err := ioutil.ReadFile(infrastructureFile)
+	if err != nil {
+		return nil, err
+	}
+	if len(fileData) > 0 {
```
abhinavdahiya, line 47: Is this check required?

pkg/operator/kube_cloud_config/transform.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	// Read, parse, and save the user provided cloud configmap
+	var cloudProviderConfigInput corev1.ConfigMap
+	if cloudProviderFile != "" {
+		fileData, err = ioutil.ReadFile(cloudProviderFile)
+		if err != nil {
+			return nil, err
+		}
+		if len(fileData) > 0 {
```
abhinavdahiya, line 61: Is this check required?

pkg/operator/kube_cloud_config/transform.go
```diff
@@ -0,0 +1,99 @@
...
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/yaml"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorclient "github.com/openshift/cluster-config-operator/pkg/operator/operatorclient"
+)
+
+func CloudConfigTransformers() map[configv1.PlatformType]cloudConfigTransformer {
```
abhinavdahiya, line 15: I think we can make this function private

pkg/operator/kube_cloud_config/transform.go
```diff
@@ -0,0 +1,99 @@
...
+	if len(platformName) == 0 {
+		platformName = clusterInfrastructure.Status.Platform
+	}
+
+	// Determine the platform specific transformer method to use
+	cloudConfigTransformers := CloudConfigTransformers()
+	cloudConfigTransformerFn, ok := cloudConfigTransformers[platformName]
+	if !ok {
+		cloudConfigTransformerFn = asIsTransformer
+	}
```
abhinavdahiya, line 83: maybe all this can be part of the `CloudConfigTransformers` that takes the infratructure object and returns the transformer.. could also be used by sync.

pkg/operator/kube_cloud_config/transform.go
```diff
@@ -0,0 +1,99 @@
+package kube_cloud_config
```
abhinavdahiya, line 1: I would think about splitting this file to 
bootstrap.go => has the handler for bootstrap transforming 
moving general transform function like `CloudConfigTransformers` to controller.go

pkg/operator/kube_cloud_config/transform.go
jhixson74: Apparently not. I did not realize it's safe to unmarshal a zero length bytes object.

pkg/cmd/render/render.go
jhixson74: Since the input and output file names are not required, the line referenced checks that if one is provided, the other must be provided as well. This line says that if they're both set, we can go ahead and call the transformer function.

pkg/operator/kube_cloud_config/transform.go
jhixson74: I should probably also write up some tests here.

pkg/cmd/render/render.go
```diff
@@ -63,6 +81,16 @@ func (r *renderOpts) Validate() error {
...
 
+	// Validate all files are specified when specifying infrastructure and configmap files
+	if len(r.clusterInfrastructureInputFile) != 0 || len(r.cloudProviderConfigOutputFile) != 0 {
+		if !(len(r.clusterInfrastructureInputFile) != 0 && len(r.cloudProviderConfigOutputFile) != 0) {
+			return fmt.Errorf("clulster-infrastructure-file and cloud-provider-config-output-file must be specified.")
+		}
+		if err := kubecloudconfig.ValidateFile(r.clusterInfrastructureInputFile); err != nil {
+			return err
+		}
+	}
```
abhinavdahiya, line 92: what if we do?

```suggestion
	// Validate all files are specified when specifying infrastructure and configmap files
	if len(r.clusterInfrastructureInputFile) !=  len(r.cloudProviderConfigOutputFile) {
		return fmt.Errorf("cluster-infrastructure-file and cloud-provider-config-output-file must be specified.")
	}
	if len(r.clusterInfrastructureInputFile) != 0 {
		kubecloudconfig.ValidateFile(r.clusterInfrastructureInputFile)
	}
```

pkg/operator/kube_cloud_config/bootstrap.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, err
+	}
+	err = yaml.Unmarshal(fileData, &clusterInfrastructure)
+	if err != nil {
+		return nil, err
+	}
+
+	// Read, parse, and save the user provided cloud configmap
+	var cloudProviderConfigInput corev1.ConfigMap
+	if cloudProviderFile != "" {
```
abhinavdahiya, line 45: nit:
```suggestion
	if len(cloudProviderFile) > 0 {
```

pkg/operator/kube_cloud_config/bootstrap.go
```diff
@@ -0,0 +1,86 @@
...
+		fileData, err = ioutil.ReadFile(cloudProviderFile)
+		if err != nil {
+			return nil, err
+		}
+		err := yaml.Unmarshal(fileData, &cloudProviderConfigInput)
```
abhinavdahiya, line 50:
```suggestion
		err = yaml.Unmarshal(fileData, &cloudProviderConfigInput)
```

abhinavdahiya:
@jhixson74 can you open a PR on the installer showing how we will use this on the bootstrap host.

you can then test these together using cluster-bot in slack
Option A:
1. `/msg @cluster-bot launch 4.6.0-0.ci,openshift/installer#<pr number>,openshift/cluster-config-operator#140 aws`
2. `/msg @cluster-bot launch 4.6.0-0.ci,openshift/installer#<pr number>,openshift/cluster-config-operator#140 azure`

Option B:
1. `/msg @cluster-bot build 4.6.0-0.ci,openshift/installer#<pr number>,openshift/cluster-config-operator#140`
2. Get the release image.
3. Ensure you have latest pull secret from api.openshift.com 
3.a run login command
3.b oc registry login
3.c make sure you add that pulll secret to the installer install-config.yaml
4. oc extract --command=openshift-install <release image>
5. Use the installer to create a cluster as you like

pkg/operator/kube_cloud_config/bootstrap.go
jhixson74: Fixed.

pkg/cmd/render/render.go
jhixson74: I don't understand what testing if the length of the input file matches the length of the output file accomplishes here since they can be different lengths, so I have left that code intact. I have added the if clause before file validation.

abhinavdahiya:
LGTM

/assign @deads2k @sttts 
Can you help review / approve this?

pkg/cmd/render/render.go
```diff
@@ -17,7 +19,12 @@ type renderOpts struct {
 	manifest genericrenderoptions.ManifestOptions
 	generic  genericrenderoptions.GenericOptions
 
+	// XXX: What is this file used for?
```
sttts, line 22: looks like unused, introduced 2y ago.

pkg/cmd/render/render.go
```diff
@@ -63,6 +81,18 @@ func (r *renderOpts) Validate() error {
 		return err
 	}
 
+	// Validate all files are specified when specifying infrastructure and configmap files
+	if len(r.clusterInfrastructureInputFile) != 0 || len(r.cloudProviderConfigOutputFile) != 0 {
```
sttts, line 85:
````
if infra, provider := len(r.clusterInfrastructureInputFile) != 0, len(r.cloudProviderConfigOutputFile); infra || provider {
  if infra && provider {
```

makes it easier to read

pkg/cmd/render/render.go
```diff
@@ -85,28 +115,42 @@ type TemplateData struct {
...
 	if err := r.generic.ApplyTo(
 		&renderConfig.FileConfig,
-
-		// TODO I don't think we have these
 		genericrenderoptions.Template{},
 		genericrenderoptions.Template{},
-
 		&renderConfig,
-		nil,
+		nil, // XXX: Can this be used to merge resources and write out config file? (map[string]resourcemerge.MergeFunc)
```
sttts, line 136: it can. But afaik we don't do that anywhere, but keep the config semantics plain and simple.

pkg/operator/kube_cloud_config/bootstrap.go
```diff
@@ -0,0 +1,86 @@
+package kube_cloud_config
```
sttts, line 1: ugly package name with underscore.

pkg/operator/kube_cloud_config/bootstrap.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// ValidateFile verifies a file exists, has content, and is a regular file
+func ValidateFile(path string) error {
+	st, err := os.Stat(path)
+	if err != nil {
+		return err
+	}
+	if !st.Mode().IsRegular() {
+		return fmt.Errorf("%s is not a regular file.", path)
```
sttts, line 22: no dots at end of errors

pkg/operator/kube_cloud_config/bootstrap.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	if st.Size() <= 0 {
+		return fmt.Errorf("%s is empty.", path)
+	}
+	return nil
+}
+
+func BootstrapTransformer(infrastructureFile string, cloudProviderFile string) ([]byte, error) {
```
sttts, line 30: go doc. What does this do?

BootstrapTransformer is  not a verb, but it should. It generates a target cloud config? Call it like that.

pkg/operator/kube_cloud_config/bootstrap.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Read, parse, and save the infrastructure object
+	var clusterInfrastructure configv1.Infrastructure
+	fileData, err := ioutil.ReadFile(infrastructureFile)
+	if err != nil {
+		return nil, err
+	}
+	err = yaml.Unmarshal(fileData, &clusterInfrastructure)
+	if err != nil {
+		return nil, err
```
sttts, line 40: give error context

pkg/operator/kube_cloud_config/bootstrap.go
```diff
@@ -0,0 +1,86 @@
...
+	// Read, parse, and save the user provided cloud configmap
+	var cloudProviderConfigInput corev1.ConfigMap
+	if len(cloudProviderFile) > 0 {
+		fileData, err = ioutil.ReadFile(cloudProviderFile)
+		if err != nil {
+			return nil, err
+		}
+		err = yaml.Unmarshal(fileData, &cloudProviderConfigInput)
+		if err != nil {
+			return nil, err
```
sttts, line 52: give error context

pkg/operator/kube_cloud_config/controller.go
```diff
@@ -154,3 +151,12 @@ func asIsTransformer(input *corev1.ConfigMap, sourceKey string, _ *configv1.Infr
 
 	return output, nil
 }
+
+// cloudConfigTransformers returns all configured cloud transformers
+func cloudConfigTransformers() map[configv1.PlatformType]cloudConfigTransformer {
```
sttts, line 156: make this a global var. No need to do this dynamically.

pkg/operator/kube_cloud_config/bootstrap.go
jhixson74: I have renamed the package to kubecloudconfig.

pkg/operator/kube_cloud_config/bootstrap.go
jhixson74: I have removed the dots.

pkg/operator/kube_cloud_config/bootstrap.go
jhixson74: Renamed to BootstrapTransform() and added comment block at top of function.

pkg/operator/kube_cloud_config/bootstrap.go
jhixson74: context added.

pkg/operator/kube_cloud_config/bootstrap.go
jhixson74: context added.

pkg/operator/kube_cloud_config/controller.go
jhixson74: Apparently, global maps in golang can't be initialized. I think it makes more sense and is cleaner. If global, then it will need to be initialized somewhere. Since it is used in a few places (and possibly more in the future), this will begin look uglier IMO. Let me know your thoughts.

pkg/cmd/render/render.go
jhixson74: This is nice. I've updated the syntax accordingly. Thanks for introducing me to this ;-)

wking:
/retitle render: generate cloud config when infrastructure and configmap are provided

To match e87cce2ad127, because the current title doesn't talk about which ConfigMap.

PR #145, opened 2020-07-24
bump to use latest openshift/api

tkashem:


sttts:
AWS quota slice took hours. Then actual install took 29min. Tests time out because everything (including quota slice step) must be finished in 4h.

sttts:
Cluster came up infinite many times, and upgrade tests worked too.

/override ci/prow/e2e-aws

PR #150, opened 2020-09-11
Bug 1878007: Add Controller to Normalize the Operator Log Level Values

vareti:
This PR adds new controller to normalize the operator log level values that are different from possible values.


pkg/operatorloglevel/loglevel_controller.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return factory.New().
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		ResyncEvery(time.Minute).
+		ToController("OperatorLogLevelNormalizer", recorder)
+}
+
+// sync runs periodically and makes sure that the operator log level values does not
+// differ from supported values. Sub-resources are ignored. Operator custom resources
```
sttts, line 42: what does "Sub-resources are ignored." mean? You mean that you ignore them in discovery?

pkg/operatorloglevel/loglevel_controller.go
```diff
@@ -0,0 +1,107 @@
...
+	operatorClient operatorv1helpers.OperatorClient,
+	recorder events.Recorder) factory.Controller {
+	c := operatorLogLevelNormalizer{
+		dynamicClient: dynamicClient,
+	}
+
+	return factory.New().
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		ResyncEvery(time.Minute).
```
sttts, line 37: would increase this. 5 min? 10 min? 30 min?

pkg/operatorloglevel/loglevel_controller.go
```diff
@@ -0,0 +1,107 @@
...
+
+// sync runs periodically and makes sure that the operator log level values does not
+// differ from supported values. Sub-resources are ignored. Operator custom resources
+// that do not have log level field are also ignored.
+func (c *operatorLogLevelNormalizer) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	// find all operator CRDs
+	apiResources, err := c.dynamicClient.Resource(schema.GroupVersionResource{
+		Group:   "operator.openshift.io",
+		Version: "v1",
+	}).List(ctx, metav1.ListOptions{})
```
sttts, line 49: does this work? I think you have to pass a concrete resource name.

There is the discovery client to get the list.

sttts: Maybe this work because the dynamic client hits the right endpoint. But better use an explicit discovery client. This here is only an accident that it works ;-)

sttts: Actually, after reading https://github.com/openshift/api/pull/736, I would just put an hard-coded slice of GVRs here. No discovery. Then also know that they are really using the underlying type with log levels we expect.

pkg/operatorloglevel/loglevel_controller.go
```diff
@@ -0,0 +1,107 @@
...
+			if !loglevel.ValidLogLevel(operatorv1.LogLevel(currentLogLevel)) {
+				crCopy := cr.DeepCopy()
+				if err := unstructured.SetNestedField(crCopy.UnstructuredContent(), string(operatorv1.Normal), operatorLogLevelPath...); err != nil {
+					klog.Warningf("failed to set log level to %s in resource %s", operatorv1.Normal, apiResource.Name)
+					continue
+				}
+
+				if _, err := c.dynamicClient.Resource(gvr).Update(ctx, crCopy, metav1.UpdateOptions{}); err != nil {
+					klog.Warningf("failed to normalize log level to %v for operator %s", operatorv1.Normal, apiResource.Name)
+				}
```
sttts, line 99: continue

pkg/operatorloglevel/loglevel_controller.go
vareti: > This here is only an accident that it works ;-)
I tested the changes yesterday and they work. I was just lucky I guess :)

>  I would just put an hard-coded slice of GVRs here
hard-coded the gvr slice now. No discovery and dynamic client.

pkg/operatorloglevel/loglevel_controller.go
```diff
@@ -0,0 +1,107 @@
...
+			Version:  "v1",
+			Resource: apiResource.Name,
+		}
+		customResources, err := c.dynamicClient.Resource(gvr).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			klog.V(4).Infof("error trying to list custom resources for %s: %v", apiResource.Name, err)
+			continue
+		}
+
+		operatorLogLevelPath := []string{"spec", "operatorLogLevel"}
```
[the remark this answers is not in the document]
vareti: added now.

pkg/operatorloglevel/loglevel_controller.go
vareti: Increased it to 5mins now.

pkg/operatorloglevel/loglevel_controller.go
```diff
@@ -0,0 +1,179 @@
...
+	return factory.New().
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		ResyncEvery(5*time.Minute).
+		ToController("OperatorLogLevelNormalizer", recorder)
+}
+
+// sync runs periodically and makes sure that the log level field values in operator
+// spec does not differ from supported values.
+func (c *operatorLogLevelNormalizer) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
sttts, line 43: please create a follow-up BZ that we can remove this controller again one release after we enabled stricter validation.

pkg/operatorloglevel/loglevel_controller.go
```diff
@@ -0,0 +1,179 @@
...
+			klog.V(4).Infof("error trying to list custom resources for %s: %v", gvr.Resource, err)
+			continue
+		}
+
+		for _, cr := range customResources.Items {
+			crCopy := cr.DeepCopy()
+			eventMsgs, needsUpdate := normalizeLogLevelField(crCopy, gvr.Resource)
+			if needsUpdate {
+				if _, err := c.dynamicClient.Resource(gvr).Update(ctx, crCopy, metav1.UpdateOptions{}); err != nil {
+					klog.Warningf("failed to normalize log level to %v for operator %s", operatorv1.Normal, gvr.Resource)
```
sttts, line 139: print the error too

pkg/operatorloglevel/loglevel_controller.go
```diff
@@ -0,0 +1,179 @@
...
+
+func normalizeLogLevelField(cr *unstructured.Unstructured,
+	resourceName string) ([]string, bool) {
+	needsUpdate := false
+	eventMsgs := []string{}
+
+	for _, logLevelFieldPath := range [][]string{{"spec", "operatorLogLevel"}, {"spec", "logLevel"}} {
+		// custom resources that do not have log level field are ignored.
+		currentLogLevel, ok, err := unstructured.NestedString(cr.UnstructuredContent(), logLevelFieldPath...)
+		if !ok || err != nil {
```
sttts, line 160: !ok is no error. This field was optional. Just do continue then.

sttts:
/lgtm
/approve

Please also create against master.

vareti:
/hold

until we can resolve the fake bump

sttts:
Why fake bump? Anything missing?

vareti:
> Why fake bump? Anything missing?

Bug mentioned in https://github.com/openshift/library-go/pull/890 is waiting for parent bug to move to verified state. Until that PR is merged, I cannot do a real bump.

markmc:
(patch manager) asked a question in the bz about the need for backporting this

sttts:
@markmc No, we won't backport. The normalization happens from 4.5 to 4.6.

markmc:
> @markmc No, we won't backport. The normalization happens from 4.5 to 4.6.

It's this PR I'm talking about - why is a backport of openshift/cluster-config-operator#151 justified? From the bz:

(patch manager) This seems like quite a significant change (wrt to risk of unanticipated side effects) relative to the value to users of an improvement in how we handle invalid log level configuration values? What am I missing?

ricardomaraschini:
Is this PR going to make 4.6 ? I have a pending one (https://github.com/openshift/api/pull/736) that we would like to land on the same version.

bparees:
i'm not sure who added the "awaiting evidence" label previously, but i'm reasserting it now per my comment in the BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1878007#c6

i don't think we should be implementing this change in 4.5, we should have implemented the migration in 4.6(before 4.6 shipped) and the validation in 4.7.

We also should not have merged the tighter validation logic into vN before we had appropriate migration logic in vN-1.

sttts:
@bparees copying my comment from the BZ:

We have the controller also in 4.6. Whether we start migrating in 4.5 or 4.6 does not make a difference.

Note that we do not even need the 4.5 controller necessarily. We would only avoid possible delays during upgrade until the config controller kicks in.

bparees:
> We have the controller also in 4.6. Whether we start migrating in 4.5 or 4.6 does not make a difference.

got it, thanks that helps

xingxingxia:
/bugzilla cc-qa
(I'm following the issue DPTP-660 to do the pre-merge verifying)

xingxingxia:
Already pre-merge verified in [1878007#c10](https://bugzilla.redhat.com/show_bug.cgi?id=1878007#c10), so per DPTP-660 adding:
/lgtm

pkg/operator/operatorloglevel/loglevel_controller.go
```diff
@@ -0,0 +1,187 @@
...
+		if loglevel.ValidLogLevel(operatorv1.LogLevel(currentLogLevel)) {
+			continue
+		}
+
+		if err := unstructured.SetNestedField(cr.UnstructuredContent(), string(operatorv1.Normal), logLevelFieldPath...); err != nil {
+			klog.Warningf("failed to set log level to %s in resource %s", operatorv1.Normal, resourceName)
+			continue
+		}
+
+		eventMsgs = append(eventMsgs, fmt.Sprintf("%q changed from %q to %q", strings.Join(logLevelFieldPath, "."), currentLogLevel, operatorv1.Normal))
```
xingxingxia, line 182: A very minor suggestion, maybe it is better to include resourceName in the event message to show which CR :)

PR #161, opened 2020-10-07
Bug 1878199: remove controller normalizing log levels in operator spec

vareti:
This PR can be merged once log level value validation is added to all the operator specs.

/hold
/assign @sttts


sttts:
> once log level value validation is added to all the operator specs.

Which are missing?

vareti:
>Which are missing?

@sttts cloud credentials and network crds do not have log level validation. vendor bump in this [commit](https://github.com/openshift/cluster-config-operator/pull/161/commits/25a4859bf0d2fa4fbbc645a76a384f9fd09a265d) takes care of that. please take a look.

/hold cancel

sttts:
> @sttts cloud credentials and network crds do not have log level validation. vendor bump in this commit takes care of that. please take a look.

You mean we missed them in 4.6 ? Don't we have to keep the controller then for another release?

vareti:
@sttts I see this on a 4.7 cluster. I dont know which operator is responsible for creating network CRD but that operator was not bumped to use the openshift/api revision validating the log level.


> oc get crd networks.operator.openshift.io -o jsonpath="{.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.logLevel}"
map[description:logLevel allows configuring the logging level of the components deployed by the operator. Currently only Kuryr SDN is affected by this setting. Please note that turning on extensive logging may affect performance. The default value is "Normal". type:string]

I was of the opinion that cluster-config-operator is responsible for creating all the CRDs but it is not. So, the vendor bump in this PR will not add any value. I will remove it.

vareti:
So, network operator does not copy its CRD from `openshift/api`. It has another [file](https://github.com/openshift/cluster-network-operator/blob/master/manifests/0000_70_cluster-network-operator_01_crd.yaml) that is not updated and that stale crd is copied to payload.

sttts:
This still have to wait another release because https://github.com/openshift/cluster-network-operator/commit/ea7eec5beddee307839511624f8bfbd7cc796459#diff-51340aea1de9eeec24aebe2404caddd9258b2e158fb1422598f336cf2d08d939 only merged for 4.7.

sttts:
/close

Replaced with https://github.com/openshift/cluster-config-operator/pull/197.

PR #175, opened 2020-12-22
bump openshift/api to support new field controlPlaneToplogy in

ravidbro:
Infrastructure

Signed-off-by: Ravid Brown <ravid@redhat.com>


deads2k:
/lgtm

You'll want to keep an eye on this to see if it goes green.

deads2k:
for future reference, https://github.com/openshift/api/commit/670ac3fc997c2f1d19b8c29ef04f70d6e3d4a59e is current HEAD of master

ravidbro:
> for future reference, [openshift/api@670ac3f](https://github.com/openshift/api/commit/670ac3fc997c2f1d19b8c29ef04f70d6e3d4a59e) is current HEAD of master

maybe I'm missing something but I think that this is the hash I put

PR #176, opened 2021-01-06
Apply label to all CRD's this operator provide

mfojtik:


Dockerfile.rhel7
```diff
@@ -4,23 +4,32 @@ COPY . .
 ENV GO_PACKAGE github.com/openshift/cluster-config-operator
 RUN make build --warn-undefined-variables
 
+# The labeler step will apply an 'openshift.io/installed-by=cluster-config-operator' label to all CRD resources this operator install.
```
sttts, line 7: use base image for the COPYies. And make the labeler code super small with its own image and step

sttts:
Sgtm

PR #191, opened 2021-04-12
Bug 1949145: Add user facing priority class

ravisantoshgudimetla:
Add user facing critical priority class to CCO. This priority class is currently in [CMO](https://github.com/openshift/cluster-monitoring-operator/blob/master/manifests/0000_50_cluster-monitoring-operator_00_0user-priority-class.yaml), however we see that it is beneficial enough to be used by multiple entities like RHOAM, so moving it CCO.

cc @smarterclayton

PR #193, opened 2021-04-15
Bug 1949145: Add missing annotations to upc

ravisantoshgudimetla:
As part of #191, I missed adding required annotations to priority class manifest. Adding them now.

cc @soltysh


soltysh (approved):
/lgtm
/approve

PR #196, opened 2021-04-21
add controller to sync aws resource tags

staebler:
Add aws_resource_tags controller. The controller will sync AWS resource tags from the spec of the infrastructure resource to the status of the resource.

/cc @deads2k


staebler:
/hold

This is using a fork of openshift/api until https://github.com/openshift/api/pull/905 merges.

pkg/operator/aws_resource_tags/controller.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+	}
+
+	var currentResourceTags []configv1.AWSResourceTag
+	if currentInfra.Status.PlatformStatus != nil &&
+		currentInfra.Status.PlatformStatus.AWS != nil {
+		currentResourceTags = currentInfra.Status.PlatformStatus.AWS.ResourceTags
+	}
+
+	if len(desiredResourceTags) == 0 && len(currentResourceTags) == 0 {
```
deads2k, line 91: seems like this case is handled by the deepequal on the next line.  is it not?

pkg/operator/aws_resource_tags/controller.go
```diff
@@ -0,0 +1,149 @@
...
+
+	if currentInfra.Status.PlatformStatus == nil {
+		currentInfra.Status.PlatformStatus = &configv1.PlatformStatus{}
+	}
+	if currentInfra.Status.PlatformStatus.AWS == nil {
+		currentInfra.Status.PlatformStatus.AWS = &configv1.AWSPlatformStatus{}
+	}
+	currentInfra.Status.PlatformStatus.AWS.ResourceTags = desiredResourceTags
+
+	_, err = c.infraClient.UpdateStatus(ctx, currentInfra, metav1.UpdateOptions{})
```
deads2k, line 107: I think an event on successful update is also warranted.

pkg/operator/aws_resource_tags/controller.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil
+	}
+
+	if currentInfra.Status.PlatformStatus == nil {
+		currentInfra.Status.PlatformStatus = &configv1.PlatformStatus{}
+	}
+	if currentInfra.Status.PlatformStatus.AWS == nil {
+		currentInfra.Status.PlatformStatus.AWS = &configv1.AWSPlatformStatus{}
+	}
+	currentInfra.Status.PlatformStatus.AWS.ResourceTags = desiredResourceTags
```
deads2k, line 105: Unless order matters, please sort the status resourcetags.

deads2k:
minor comments, but I think the API and additional validation make sense and provide clear indications of what can and cannot be done.

PR #200, opened 2021-04-27
Bug 1954152: vendor: bump to the latest openshift/api

staebler:
Upgrade github.com/openshift/api to latest 4.7 version.

This is needed so that the migration_platform_status does not clear out the new AWS resourceTags field, which is unknown without the api bump.

This is a manual cherry-pick of #199.


staebler:
> /hold for [openshift/api#914](https://github.com/openshift/api/pull/914)

PR has merged.
/hold cancel

deads2k:
confirmed it points to api/4.7

/lgtm

PR #203, opened 2021-05-27
bump(*): openshift/api to pickup updated CRD's

mfojtik:


mfojtik:
(this PR does not require valid-bug label as it is just updating CRD's that has new annotation)

PR #209, opened 2021-07-12
bump: get updated operator API content

deads2k:
pick up https://github.com/openshift/api/pull/966


deads2k:
purely generated PR

PR #213, opened 2021-07-28
Bug 1984635: use new default leader election values to handle SNO environments

aojea:
It also removes klog v1 dependency.

I wrongly updated openshift-api instead of library-go in previous PR

https://github.com/openshift/cluster-config-operator/pull/211


vendor/github.com/openshift/library-go/pkg/config/leaderelection/leaderelection.go
```diff
@@ -84,14 +84,28 @@ func ToConfigMapLeaderElection(clientConfig *rest.Config, config configv1.Leader
...
+	// We want a single set of guidance for nearly every lease in openshift.  If you're special, we'll let you know.
+	// 1. clock skew tolerance is leaseDuration-renewDeadline == 30s
+	// 2. kube-apiserver downtime tolerance is == 78s
+	//      lastRetry=floor(renewDeadline/retryPeriod)*retryPeriod == 104
+	//      downtimeTolerance = lastRetry-retryPeriod == 78s
+	// 3. worst non-graceful lease acquisition is leaseDuration+retryPeriod == 163s
+	// 4. worst graceful lease acquisition is retryPeriod == 26s
 	if ret.LeaseDuration.Duration == 0 {
-		ret.LeaseDuration.Duration = 60 * time.Second
+		ret.LeaseDuration.Duration = 137 * time.Second
```
aojea, line 97: new leader election values

PR #227, opened 2021-12-10
[WIP] feat: bumped library-go version

eggfoobar:
Bumping the `library-go` dependency to latest, this should take advantage of leader election changes for SNO clusters proposed in this [library-go PR](https://github.com/openshift/library-go/pull/1242). 

Changes:
- updated library-go to use latest 
- updated method call to pass context to match updated method signature

Signed-off-by: ehila <ehila@redhat.com>


cmd/cluster-config-operator/main.go
```diff
@@ -1,37 +1,21 @@
...
 	"github.com/openshift/cluster-config-operator/pkg/cmd/operator"
 	"github.com/openshift/cluster-config-operator/pkg/cmd/render"
 	"github.com/openshift/cluster-config-operator/pkg/version"
 )
 
 func main() {
-	rand.Seed(time.Now().UTC().UnixNano())
-
-	pflag.CommandLine.SetNormalizeFunc(utilflag.WordSepNormalizeFunc)
-	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
```
deads2k, line 25: I don't see component-base/cli adding the golang flag set to catch libraries that make use of them.

cmd/cluster-config-operator/main.go
```diff
@@ -1,37 +1,21 @@
...
-
-	logs.InitLogs()
```
deads2k, line 27: the new library appears to lose log initialization from before the flag parse.

cmd/cluster-config-operator/main.go
```diff
@@ -1,37 +1,21 @@
...
-	defer logs.FlushLogs()
-
 	command := NewOperatorCommand()
-	if err := command.Execute(); err != nil {
-		fmt.Fprintf(os.Stderr, "%v\n", err)
```
deads2k, line 32: am I misreading or does the new library dump this along with standard log output (which might be redirected incorrectly to stderr) instead of stderr

cmd/cluster-config-operator/main.go
eggfoobar: That should be done during this init function https://github.com/kubernetes/component-base/blob/v0.23.0/logs/config.go#L49

cmd/cluster-config-operator/main.go
eggfoobar: Right, I might have misunderstood if that's a desired behavior, given this line comment https://github.com/kubernetes/component-base/blob/master/logs/logs.go#L170 in the `InitLogs()`

cmd/cluster-config-operator/main.go
eggfoobar: Hmm, as I understood it the core logic doesn't change much, except that they differentiate when it's a flag parse error versus a runtime error and use the corresponding method for output, in this [line](https://github.com/kubernetes/component-base/blob/v0.23.0/cli/run.go#L105)

eggfoobar:
/retitle Bug 2034484: feat: bumped library-go version
This PR bumps to latest library-go to utilize changes for performance gains on SNO DU clusters

cmd/cluster-config-operator/main.go
eggfoobar: reverted back start command to simplify PR

cmd/cluster-config-operator/main.go
eggfoobar: reverted back start command to simplify PR

cmd/cluster-config-operator/main.go
eggfoobar: reverted back start command to simplify PR

cmd/cluster-config-operator/main.go
```diff
@@ -25,6 +25,7 @@ func main() {
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 
 	logs.InitLogs()
+	logs.AddFlags(pflag.CommandLine)
```
sttts, line 28: don't mix bumps with changes

eggfoobar: Hey @sttts , after the bump of `component-base` the way the flags get initialized from `func init` in the logger changed in this [commit](https://github.com/kubernetes/component-base/commit/0c68d203462388c5319702773d799232d01def68) so this is explicitly binding the flags to make sure we get the same behavior, otherwise the logging flags don't get added to `pflag` like they were before.

eggfoobar: Actually, I missed that this [PR](https://github.com/openshift/cluster-config-operator/pull/229) already updated component base, if it's still desired to have the logger flags be part of the flag parsing, I'll open up a new PR to add that line.

eggfoobar:
Hey @deads2k when you get a chance, this should be all set for another look.

PR #244, opened 2022-04-13
Reorder the empty Node CR resource to ensure it is applied after the Node CRD

bparees:
we think the ordering was causing upgrade failures like:

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.11-e2e-gcp-upgrade/1513978064520876032

because the CVO was trying to apply the CR before the CRD was applied, and got stuck.

more details in slack thread:
https://coreos.slack.com/archives/C01CQA76KMX/p1649814384916179


bparees:
@swghosh fyi

wking (approved):
Fixing the backwards order in:

```console
$ oc adm release extract --to manifests registry.ci.openshift.org/ocp/release:4.11.0-0.ci-2022-04-12-202956
$ grep -r 'kind: Node' manifests
manifests/0000_05_config-operator_02_node.cr.yaml:kind: Node
manifests/0000_10_config-operator_01_node.crd.yaml:    kind: Node
```

With this change, the CRD will sort before the CR (which depends on the CRD).  We probably want to do the rest of the 05 CRs as well, but we can do that in follow-up work.

/lgtm

swghosh:
Thanks @bparees!

Cc: @rphillips @harche

PR #251, opened 2022-05-04
bump openshift/api, client-go

sairameshv:
Bumped the openshift/api, client-go dependencies which include the removal of the worker latency profile related status from the config node custom resource definition.


soltysh (approved):
/lgtm
/approve

PR #268, opened 2022-10-12
NE-975: Update openshift/api for updated ingress config loadBalancer fields

gcs278:
For [NE-975](https://issues.redhat.com/browse/NE-975), we need the `ingress.config.openshift.io` CRD to have the loadBalancer field for specifying loadBalancer type upon installation. See https://github.com/openshift/api/pull/1209, https://github.com/openshift/cluster-ingress-operator/pull/837, and https://github.com/openshift/installer/pull/6478

- `go.mod`: Bumped openshift/api to `v0.0.0-20220927015011-24ee13c10f3b` and also bumped k8s modules to `0.25.2` due to dependency issues (`make verify` was failing with K8S module conflicts). I also weirdly had to add `k8s.io/apiserver v0.25.2` to required, `go mod tidy` and `go mod vendor` in order to resolve apiserver related issues. But then `go mod tidy` removed it from the `require` section, but the indirect dependency is now 0.25.0. Not sure what was going on.

CC: @miheer @Miciah


vendor/github.com/openshift/api/config/v1/0000_10_config-operator_01_ingress.crd.yaml
```diff
@@ -81,6 +81,46 @@ spec:
                 domain:
                   description: "domain is used to generate a default host name for a route when the route's host name is empty. The generated host name will follow this pattern: \"<route-name>.<route-namespace>.<domain>\". \n It is also used as the default wildcard domain suffix for ingress. The default ingresscontroller domain will follow this pattern: \"*.<domain>\". \n Once set, changing domain is not currently supported."
                   type: string
+                loadbalancer:
```
Miciah, line 84: We're missing https://github.com/openshift/api/commit/24ee13c10f3bf7c3320392ba687a5a42bb3c04c4.

gcs278: Whoops, done

gcs278:
/hold
issues with K8S modules, I think I need to bump

PR #279, opened 2023-02-06
CFE-601 : Update openshift/api package version

bharath-b-rh:
https://github.com/openshift/enhancements/pull/1191 proposes support for Azure tags support in OCP and requires changes in openshift/api Infrastructure CRD and is used by openshift/installer for updating the user defined tags, which is made available for installer by cluster-config-operator, which is included in the container image from the openshift/api package and requires the package to be updated to have the latest CRD.


deads2k:
/hold

covered by https://github.com/openshift/cluster-config-operator/pull/278 which I've just labelled.

PR #294, opened 2023-04-24
update ordering and featuregates

deads2k:
This PR updates the ordering of the operator to come first.  This was the original intent in https://github.com/openshift/enhancements/blob/5180e5873f5d5aca4fe4678882602a0c7f083a63/dev-guide/cluster-version-operator/dev/upgrades.md?plain=1#L49 .  This will make the upgrade ordering successful for detecting featuregates.

It also updates the API to eliminate a failing featuregate that prevents switching kas-o.

/assign @benluddy


deads2k:
needed to unstick the CCM unrevert

PR #295, opened 2023-04-26
update render to handle directories of manifests

deads2k:
/hold

per @JoelSpeed request


deads2k:
/hold

holding until the library-go is merged.  We'll use the same reading bits to feed operators

JoelSpeed:
Looks like one of the API bumps and type changes hasn't been accounted for

PR #297, opened 2023-04-28
specify all known featureGates in disabled

deads2k:
This will allow panic-ing on unknown featuregates in consumer code.  CustomNoUpgrade becomes even harder because it must specify everything, but it's always been hard.

/assign @JoelSpeed

### DIFF
--- a/pkg/operator/featuregates/featuregate_controller.go
+++ b/pkg/operator/featuregates/featuregate_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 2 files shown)


JoelSpeed:
> CustomNoUpgrade becomes even harder because it must specify everything, but it's always been hard.

Is that true? Not how I read the code or the tests, looked to me like we were just adding everything known to disabled, and then the only enabled stuff is stuff that is explicitly enabled by CustomNoUpgrade

deads2k:
nearly there for the extgernal ccm

PR #300, opened 2023-05-02
update openshift/api

EmilienM:
That will include the latest openshift/api with external LB GA support
for OSP.


EmilienM:
/assigh deads2k

thanks

EmilienM:
/assigh deads2k

thanks

JoelSpeed:
This is covered by #301 which is tagged and includes newer updates, won't need this once 301 merges
/close

PR #303, opened 2023-05-09
OCPBUGS-6266: Rename config-operator_00_namespace run level to 00

eranco74:
During cluster installation, bootkube fails to create several resources because the "openshift-config" namespace hasn't been created yet. e.g.
Jan 12 17:37:09 master1 cluster-bootstrap[5156]: Failed to create "0000_00_cluster-version-operator_01_adminack_configmap.yaml" configmaps.v1./admin-acks -n openshift-config: namespaces "openshift-config" not found


JoelSpeed:
/lgtm

Seems reasonable to move this up

PR #304, opened 2023-05-15
Update library-go dependency

JoelSpeed:
Bump to include https://github.com/openshift/library-go/pull/1528 to fix the current feature gate bootstrapping failures.

/hold need to fixup the go.mod once we've proven this works

PR #305, opened 2023-05-15
Update library-go dependency

deads2k:
https://github.com/openshift/cluster-config-operator/pull/304 only with the right go mod.


deads2k:
pulled labels from https://github.com/openshift/cluster-config-operator/pull/304

deads2k:
old installer won't work. 

/override ci/prow/e2e-upgrade

PR #309, opened 2023-05-19
remove dead flag for file

deads2k:
The manifest directory is used instead.

/hold

holding until hypershift and @sjenning are ready.

/assign @JoelSpeed


JoelSpeed:
/hold cancel
/lgtm

Flag was removed from HyperShift in https://github.com/openshift/hypershift/pull/2585/

PR #314, opened 2023-06-01
Bump to latest openshift/api

patrickdillon:
Pulls in changes for AWS private hosted zone feature

PR #316, opened 2023-06-02
update openshift/api for types and gates

deads2k:


patrickdillon:
> /hold [openshift/api#1480](https://github.com/openshift/api/pull/1480)

Feel free to drop the hold, just wanted to note that I caught this and perhaps it would be easier to fix then revendor

deads2k:
/hold cancel

open another with the new update.

deads2k:
adds featuregates to allow features to isolate.

patrickdillon:
> /hold cancel
> 
> open another with the new update.

Reopened https://github.com/openshift/cluster-config-operator/pull/314

PR #317, opened 2023-06-05
Update openshift/api to disable EventedPLEG featuregate in techpreview

JoelSpeed:
This currently also pulls in the changes from #314 

See https://github.com/openshift/api/pull/1483 for context on why this is being disabled temporarily


sinnykumari:
Needs valid BZ

deads2k:
freeing up featuregates.

deads2k:
> For context, if you observe [these earlier test failures](https://pr-payload-tests.ci.openshift.org/runs/ci/25a8ecd0-0052-11ee-90b3-9a7ef2d2135a-00) on tech preview, these have `EventedPLEG=true` in the kubelet config, which enables the featuregate on the kubelet level.
> 
> Some symptoms that we are seeing are [pods stuck in leader election](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/openshift-machine-config-operator-3688-ci-4.14-e2e-aws-sdn-techpreview/1664179697166061568/artifacts/e2e-aws-sdn-techpreview/gather-extra/artifacts/pods/openshift-machine-config-operator_machine-config-controller-5564984d58-k9x9m_machine-config-controller.log) and pods with [bind already in use](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/openshift-machine-config-operator-3688-ci-4.14-e2e-aws-sdn-techpreview/1664179697166061568/artifacts/e2e-aws-sdn-techpreview/gather-extra/artifacts/pods/openshift-machine-config-operator_machine-config-controller-5564984d58-k9x9m_oauth-proxy.log) issues.

@mrunalp before the eventedPLEG goes in again, we should have e2e tests that fail on this condition so we get a clear fix signal on payload test.

JoelSpeed:
> I just want to highlight here that Evented PLEG was never enabled in kubelet config because the wiring to do that is still https://github.com/openshift/api/pull/1458.
Only the Tech preview feature gate https://github.com/openshift/api/pull/1471 to enable the evented pleg was added which wouldn't interact with MCO code in anyway to enable the feature in the Kubelet Config.

MCO, in master today, takes feature gates from the openshift/api repository and passes them directly through to the kubelet. So as soon as something exists in openshift/api, and is revendored into MCO, it gets passed through to Kubelet, no additional wiring needed.

In the future it will come from cluster-config-operator rather than being vendored, but will still be passed through.

Both myself and @sergiordlr grabbed kubelet configs from our clusters and could see `EventedPLEG=true` inside the kubelet config. [(example in slack)](https://redhat-internal.slack.com/archives/C02CZNQHGN8/p1685718713322659?thread_ts=1685696114.233689&cid=C02CZNQHGN8)

I think we need to have a chat about how the feature gates stuff works as gating an upstream feature gate behind a configuration in the API is not necessarily straightforward and needs some work in MCO to support it, if that's something we want to support.

harche:
@JoelSpeed thanks for that clarification. 

When I logged into your system the evented pleg was not in use, it was generic pleg that was updating the cache. This was expected because, even if you enable evented pleg in the kubelet but not in crio, the kubelet will automatically fall back to the generic pleg. 

Anyway, we would look into it more closely from node side. Thanks for your help.

PR #320, opened 2023-06-14
OCPBUGS-12767: Add CustomNoUpgrade CRD versions to payload

JoelSpeed:
Currently, CustomNoUpgrade clusters have no CRD management because there is no CRD matching the correct feature set. This PR updates openshift API to bring in new types that will work for CustomNoUpgrade clusters.


wking (approved):
/lgtm

PR #324, opened 2023-06-28
stomp the latencysensitive featureset for equivalent default

deads2k:
This let's us avoid dead-ending clusters.

PR #325, opened 2023-06-28
OCPBUGS-15877: go upgradeable=false when latencysensitive is used and not corrected

deads2k:
There are relatively few clusters even potentially impacted and we further limit that by automatically setting the equivalent featureset of "" in https://github.com/openshift/cluster-config-operator/pull/324 . This blocks upgrades if the value is not "".

### DIFF
--- a/pkg/operator/featureupgradablecontroller/OWNERS
+++ b/pkg/operator/featureupgradablecontroller/OWNERS
@@ -1,0 +1,9 @@
+# https://github.com/openshift/installer/blob/75738a342c1973121eedda7d91096d21c19194c9/OWNERS_ALIASES#L47-L50
+
+reviewers:
+- deads2k
+- joelspeed
+approvers:
+# these are the api-approvers from openshift/api
+- deads2k
+- joelspeed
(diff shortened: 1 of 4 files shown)


JoelSpeed:
Changes look good, but you'll need to add a new exception to the "should start all core operators test" else it will break all TPNU jobs if we merge this
```
{  fail [github.com/openshift/origin/test/extended/operators/operators.go:114]: Some cluster operators are not ready: config-operator (Upgradeable=False FeatureGates_RestrictedFeatureGates_TechPreviewNoUpgrade: FeatureGatesUpgradeable: "TechPreviewNoUpgrade" does not allow updates)
Ginkgo exit error 1: exit with code 1}
```

dinhxuanvu:
> Changes look good, but you'll need to add a new exception to the "should start all core operators test" else it will break all TPNU jobs if we merge this
> 
> ```
> {  fail [github.com/openshift/origin/test/extended/operators/operators.go:114]: Some cluster operators are not ready: config-operator (Upgradeable=False FeatureGates_RestrictedFeatureGates_TechPreviewNoUpgrade: FeatureGatesUpgradeable: "TechPreviewNoUpgrade" does not allow updates)
> Ginkgo exit error 1: exit with code 1}
> ```

It looks like the test meant to ignore certain conditions but the current [check](https://github.com/openshift/origin/blob/master/test/extended/operators/operators.go#L93) doesn't with what David introduced in this PR. The test can be changed to address this problem.

PR #326, opened 2023-07-04
Enable feature gate for externalCertificate on Route API

thejasn:
Follow up to https://github.com/openshift/api/pull/1387


JoelSpeed:
/lgtm

This should be inert at this stage so safe to merge

PR #329, opened 2023-07-13
OCPBUGS-16164: retire LatencySensitive featureset

deads2k:
For multiple years this has been the equivalent of "".  Make these consistent because in 4.13 the featureset handling added to the CVO didn't account for this and its unnecessary.  It's more practical to migrate a few existing clusters than modify all producers.


JoelSpeed:
Units appear to have failed on related tests but don't have logs captured, so not sure what the actual issue is

PR #331, opened 2023-07-18
OCPCLOUD-2010: Remove feature gate for external platform

adriengentil:
Re-vendor openshift/api in order to move external platform feature out
of the feature gate.

x-ref: https://github.com/openshift/api/pull/1530


JoelSpeed:
Are there any jobs that can be run with this PR to prove the feature is working as expected?

elmiko:
@JoelSpeed i don't think we have a ci job that exercises the external ccm in place yet

adriengentil:
/retitle OCPCLOUD-2010: Remove feature gate for external platform

PR #332, opened 2023-07-19
Bump api to update route externalCertificate

thejasn:
Follow up for: https://github.com/openshift/api/pull/1531


JoelSpeed:
/close

#331 is more up to date and will bring this change in

PR #347, opened 2023-08-22
add featuregates submodule

deads2k:
/hold

requires https://github.com/openshift/api/pull/1566, but the usage of updating multiple vendor directories may be worse than the disease.

This is coming up because two unrelated changes ended up reverted in order to pull back out a "bad" FeatureGate.

Adding the serial tests should mitigate the risk in the future, so we could decide to defer this work until we have a second incident.

/hold

Looking at the go mod vendor patterns for multiple modules in one repo, I'm not excited, so let's look before we leap.


featuregates/go.mod
```diff
@@ -0,0 +1,28 @@
+module github.com/openshift/cluster-config-operator/featuregates
```
JoelSpeed, line 1: Are we doing this to avoid Go mod hell for other repos because of the many deps pulled in by the main module? Or is this to avoid circular module deps?

If it's the first, I think lets try and stay away from this, in theory Go's dep resolution should be able to pick up only the deps you need for the packages you import from a module, and should be smart enough to not cause issues

JoelSpeed (commented):
If we can at all avoid the second go.mod, I think we should, having a separate module will likely just cause pain and confusion.

Whenever someone updates the feature gates, they then have to make sure to do `go mod vendor` in two places in this repo to actually get the result they want

deads2k:
cure is (currently) worse than the disease.

PR #349, opened 2023-08-23
read featuregates from openshift/api

deads2k:
https://github.com/openshift/api/pull/1567 is needed first, but this is how it would be used.

/hold until 4.15

### DIFF
--- a/manifests/0000_10_config-operator_07_deployment.yaml
+++ b/manifests/0000_10_config-operator_07_deployment.yaml
@@ -36,6 +36,26 @@ spec:
         secret:
           secretName: config-operator-serving-cert
           optional: true
+      initContainers:
+        - name: openshift-api
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop: ["ALL"]
+          terminationMessagePolicy: FallbackToLogsOnError
+          image: quay.io/openshift/origin-cluster-config-api:v4.0
+          imagePullPolicy: IfNotPresent
+          command:
+            - write-available-featuresets
+            - asset-output-dir=/available-featuregates
+            - payload-version=$(OPERATOR_IMAGE_VERSION)
+          resources:
+            requests:
+              memory: 50Mi
+              cpu: 10m
+          env:
+            - name: OPERATOR_IMAGE_VERSION
+              value: "0.0.1-snapshot"
       containers:
       - name: openshift-config-operator
         securityContext:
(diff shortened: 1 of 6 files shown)


JoelSpeed (commented):
LGTM pending vendor changes

Are there any changes needed to the CCO render executed by the installer to go along with this?

deads2k:
> write-available-featuresets: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by write-available-featuresets)

> write-available-featuresets: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by write-available-featuresets)

from the init container for openshift/api.  Not sure what I did wrong there.

manifests/0000_10_config-operator_07_deployment.yaml
```diff
@@ -36,6 +36,26 @@ spec:
...
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop: ["ALL"]
+          terminationMessagePolicy: FallbackToLogsOnError
+          image: quay.io/openshift/origin-cluster-config-api:v4.0
+          imagePullPolicy: IfNotPresent
+          command:
+            - write-available-featuresets
+            - asset-output-dir=/available-featuregates
```
JoelSpeed, line 50: Do you want to use an empty dir for this? And use that to share the manifests between the two containers?

Currently the E2E is failing on unable to `mkdir`

deads2k:
Looks like it works once we match the openshift/api.

This means we *could* merge, but we'd end up jamming openshift/api PRs until we managed to also complete https://github.com/openshift/installer/pull/7666, so we should wait.

deads2k:
some infrastructure appears down

/retest

JoelSpeed:
Failure in serial looks to be unrelated, will run once more to see if it goes green
/retest

JoelSpeed (commented):
/lgtm

deads2k:
https://github.com/openshift/installer/pull/7666 switching the installer rendering to use openshift/api has tested green and is ready to make the switch

https://github.com/openshift/api/pull/1666 and https://github.com/openshift/cluster-config-operator/pull/379 have clusterbot tested together fine. They hand off the CVO management of CRDs (and others) from cluster-config-operator to api image.

https://github.com/openshift/cluster-config-operator/pull/349 switches the featuregate setting from using the vendored copy from to using openshift/api.  It tested green until it openshift/api updated without a (yet) merged openshift/cluster-config-operator update, so the machine-config-operator mismatched until they rematch. But in combination with the rendering change linked above they will match ok.

Manually merging to get a batch.

PR #370, opened 2023-10-30
add image reference for openshift/api

deads2k:


JoelSpeed:
/hold

The API image is not yet marked for the payload, depends on https://github.com/openshift/api/pull/1642

PR #379, opened 2023-11-15
yield CRDs and empty resource rendering to api imge

deads2k:
/hold

must merge together with https://github.com/openshift/api/pull/1666

While the vendored level in cluster-config-operator matches, these can show green simultaneously.  When that happens, we need to manually merge both to avoid future conflicts.


deads2k:
testing combined with https://github.com/openshift/api/pull/1666 using clusterbot

deads2k:
launch openshift/cluster-config-operator#379,openshift/api#1666 worked

deads2k:
https://github.com/openshift/installer/pull/7666 switching the installer rendering to use openshift/api has tested green and is ready to make the switch

https://github.com/openshift/api/pull/1666 and https://github.com/openshift/cluster-config-operator/pull/379 have clusterbot tested together fine. They hand off the CVO management of CRDs (and others) from cluster-config-operator to api image.

https://github.com/openshift/cluster-config-operator/pull/349 switches the featuregate setting from using the vendored copy from to using openshift/api.  It tested green until it openshift/api updated without a (yet) merged openshift/cluster-config-operator update, so the machine-config-operator mismatched until they rematch. But in combination with the rendering change linked above they will match ok.

Manually merging to get a batch.

PR #396, opened 2024-01-09
OCPBUGS-26441: disable psa

ibihim:
# What

Disable the enforcement of Pod Security Standards by bumping openshift/api.

As a side-effect, I am pulling in as well:

- I need to bump client-go as well.

# Why

- There are to many pod security violations on prod clusters.
- The newest release-4.15 openshift/api contains the other changes.


JoelSpeed:
This PR isn't need. For 4.15 onwards feature gate changes are reflected into the cluster directly from the openshift/api repo. If you check the E2E jobs from the API Pr you merged, you can see the PSA feature is disabled in the feature gate status already

PR #405, opened 2024-02-02
STOR-1700: Bump(openshift/api): to get new VolumeGroupSnapshot API

bertinatto:
CC @openshift/storage


JoelSpeed:
/lgtm

@bertinatto Just as an FYI, if this change was intended to bring the new CRDs into payload, that's no longer required. Bumps here are only needed now if the config operator itself would be interacting with the resource

PR #411, opened 2024-03-06
NO-JIRA: add check to ensure clusterprofile matches for featuregates

deads2k:
This should be a safe first step towards adding clusterprofiles to openshfit/api's payload manifest.

/assign @JoelSpeed

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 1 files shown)


deads2k:
testing with clusterbot `launch 4.16,openshift/api#1788,openshift/cluster-config-operator#411 aws`

pkg/operator/starter.go
```diff
@@ -214,10 +215,27 @@ func (o *OperatorOptions) RunOperator(ctx context.Context, controllerContext *co
 	return nil
 }
 
-func (o *OperatorOptions) getFeatureGateMappingFromDisk() (map[configv1.FeatureSet]*configv1.FeatureGateEnabledDisabled, error) {
+func (o *OperatorOptions) getFeatureGateMappingFromDisk(ctx context.Context, configClient configv1client.Interface) (map[configv1.FeatureSet]*configv1.FeatureGateEnabledDisabled, error) {
+	// TODO get the cluster profile in a better way, this isn't strictly correct
```
JoelSpeed, line 219: How does CVO work it out? Is this the same mechanism?

deads2k: > How does CVO work it out? Is this the same mechanism?

CVO gets it via an env var that no other pod gets.

JoelSpeed:
/lgtm

Assuming all tests pass

/hold in case timed release required

deads2k:
ExternalOIDC is properly set so the hypershift default is contained as expected.

/hold cancel

PR #416, opened 2024-04-12
WIP: Test MetricsServer FeatureGate to default

slashpai:
/hold

Not intended to review/merge. This is only to test my change https://github.com/openshift/api/pull/1851 works as expected in https://github.com/openshift/cluster-monitoring-operator/pull/2317

cc: @simonpasquier


simonpasquier:
The Go build fails.

slashpai:
checked with API team, We no-longer needs vendoring api changes here
/close

PR #423, opened 2024-07-28
CFE-921: Update openshift/api package to latest version

bharath-b-rh:
PR is for updating the openshift/api package to the latest version where GCPLabelsTags featuregate is removed.


go.mod
```diff
@@ -13,11 +15,11 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.4
 	gopkg.in/gcfg.v1 v1.2.0
-	k8s.io/api v0.29.0
-	k8s.io/apimachinery v0.29.0
+	k8s.io/api v0.30.1
+	k8s.io/apimachinery v0.30.1
 	k8s.io/client-go v0.29.0
```
JoelSpeed, line 20: Could you please make sure that all of the k8s.io packages that are on 0.29.0 are bumped to the same level, all should move to 0.30.1

PR #427, opened 2024-09-13
OCPNODE-2399: Bump openshift/api

sairameshv:
Bump openshift/api to include the nodes.config status field


sairameshv:
/close 
As the [MCO PR](https://github.com/openshift/machine-config-operator/pull/4582) already contains the new API field

PR #432, opened 2025-03-21
OCPNODE-3029: WIP: add support for required mininum kubelet version

haircommander:


pkg/operator/featuregates/featuregate_controller_test.go
```diff
@@ -522,10 +558,89 @@ func TestFeatureGateController_sync(t *testing.T) {
...
+			cvoVersions: []string{"current-version", "prior-version"},
+			firstFeatureGate: featureGateBuilder().
+				withFeatureSet(configv1.Default).
+				statusEnabled("current-version", "One", "Two", "Seven").
+				statusDisabled("current-version", "Apple", "Banana").
+				toFeatureGate(),
+			fields: fields{
+				processVersion: "current-version",
+			},
+			minimumKubeletVersion: "1.30.0",
```
JoelSpeed, line 620: I was expecting seven to be enabled in this case? The min here is 1.30 which matches the version in the gate itself?

JoelSpeed: 🤦‍♂️ I can't read today apparently, all good

PR #439, opened 2025-04-28
OCPEDGE-1833: feat: bump api dep to pull in change in DualReplica

eggfoobar:
DualReplica went down to DevPreview from TechPreview, updating api dep to reflect this change


eggfoobar:
/cc @JoelSpeed 

This should pull in the change we made for downgrading DualReplica to DevPreview only.

PR #440, opened 2025-06-09
OCPBUGS-57187: featuregates: always complete featuregates when feature set is CustomNoUpgrade

everettraven:
In OCPBUGS-57187 it was identified that setting only the `.spec.featureSet` field to `CustomNoUpgrade` and not configuring a value for the `.spec.customNoUpgrade` field in the `FeatureGate` API resulted in an empty enabled/disabled set of feature gates, leading to panics in various platform operators.

This PR ensures that even in this edge case, we always fill the list of enabled/disabled feature gates with the default set.
It also adds a unit test to capture this scenario.


JoelSpeed:
/approve

/hold for testing

Changes seem reasonable and the units look good, but lets make sure we E2E this manually before merging

PR #441, opened 2025-06-10
[ART-13124] append .0 to go.mod version

ashwindasr:


JoelSpeed:
@ashwindasr What prevents us breaking this in the future?

ashwindasr:
@JoelSpeed for now, due to this deadline, we need to make sure that all builds pass in hermeto in Brew, since its going to be decommissioned on June 20th (ref [announcement](https://redhat-internal.slack.com/archives/C02AX10EQJW/p1748962740499999))

On konflux, we have the ability to add the .0 behind the scenes, which is not possible for us to do in Brew/Distgit. And we will be moving to Konflux in the next couple of months.

PR #495, opened 2026-06-26
OCPEDGE-2747: feat: add topology transition controller for day-2 SNO to HA transitions

jeff-roche:
## Summary

- Vendors updated openshift/api (Infrastructure spec.controlPlaneTopology, MutableTopology feature gate) and openshift/client-go (operator clientset, informers, listers for Etcd CR access)
- Adds a new `TopologyTransitionController` gated behind the `MutableTopology` feature gate that manages day-2 control plane topology transitions (initial: SNO → HA Compact on platform type None)
- Wires the controller into `RunOperator` with dedicated informer factories for the Etcd operator CR and openshift-etcd ConfigMaps

### Controller behavior

1. **Preflight validation** — verifies control plane node count, readiness, schedulability, etcd quorum, voting members, and absence of dedicated workers before allowing a transition
2. **Condition management** — sets `TopologyTransitionControllerProgressing` and `TopologyTransitionControllerUpgradeable` conditions to block cluster upgrades during transitions
3. **Status update** — applies the topology change to Infrastructure status with conflict retry and spec-change detection
4. **Reconciliation monitoring** — waits for all ClusterOperators to stabilize (Available=True, Progressing=False, Degraded=False) with a 5-minute soak timer before re-enabling upgrades
5. **Crash recovery** — handles partial failures where operator conditions were set but the Infrastructure status update didn't complete

## Test plan

- [ ] Unit tests pass: `go test -v -count=1 ./pkg/operator/topology_transition_controller/...` (46 tests covering controller sync, preflight validators, reconciliation checks, soak timer, crash recovery, and spec-change-during-retry)
- [ ] `make build` succeeds
- [ ] `make verify` passes (gofmt, govet, deps, bindata)
- [ ] E2E: deploy on a SNO cluster with MutableTopology feature gate enabled, scale to 3 control plane nodes, set `spec.controlPlaneTopology: HighlyAvailable`, verify transition completes and upgrades are re-enabled

Assisted-by: Claude <noreply@anthropic.com>


jeff-roche:
### Testing Update

Utilizing this PR and some automation scripts our team has, I was able to successfully transition an SNO platform:none cluster to HA Compact. The only real snafu was I had to manually scale ingress because ingress get's it's replica count at install time and does not update post-install.

cc @JoelSpeed

JoelSpeed:
> The only real snafu was I had to manually scale ingress because ingress get's it's replica count at install time and does not update post-install.

Is that something you think we should try to change?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/cluster-config-operator
 
-go 1.25.0
+go 1.25.5
```
JoelSpeed: Do we have to bump this patch version? IIUC it's best practice to leave this at .0

pkg/operator/topology_transition_controller/types.go
```diff
@@ -0,0 +1,99 @@
...
+// buildSupportedTransitions returns the set of permitted topology transitions
+// with validators wired to the necessary tools for accessing cluster information.
+func buildSupportedTransitions(nodeLister corev1listers.NodeLister, etcdConfigMapLister corev1listers.ConfigMapNamespaceLister, etcdLister operatorv1listers.EtcdLister) []TransitionDescriptor {
+	return []TransitionDescriptor{
+		// SNO to HA Compact on platformType: None
+		{
+			From: configv1.InfrastructureStatus{
+				ControlPlaneTopology:   configv1.SingleReplicaTopologyMode,
+				InfrastructureTopology: configv1.SingleReplicaTopologyMode,
+				PlatformStatus:         &configv1.PlatformStatus{Type: configv1.NonePlatformType},
```
JoelSpeed, line 50: When we want to expand this to additional platforms, what will that look like here?

jaypoulz:
@JoelSpeed i raised this to the ingress team earlier this year and was told that's a bug. IIRC, it happens because the ingress hard-codes the number of replicas in the ingress-controller spec when it's first created. The easy workaround is to delete the the ingress controller definition. The better solution is to have ingress reconcile changes to topology. I don't think logic for handling that belongs in CCO.

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,306 @@
...
+const (
+	transitionCondition  = "TopologyTransitionControllerProgressing"
+	upgradeableCondition = "TopologyTransitionControllerUpgradeable"
+
+	reasonTopologyTransitionInProgress = "TopologyTransitionInProgress"
+
+	// minReconciliationSoakTime is the minimum time to wait after a transition
+	// starts before accepting reconciliation checks as passing. This prevents
+	// premature completion when downstream operators haven't started progressing yet.
+	minReconciliationSoakTime = 5 * time.Minute
```
JoelSpeed, line 34: How did you choose 5 minutes?

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,306 @@
...
+) factory.Controller {
+	c := &TopologyTransitionController{
+		operatorClient: operatorClient,
+		infraLister:    infraLister,
+		infraClient:    infraClient.Infrastructures(),
+		preflightChecks: []TransitionValidatorFunc{
+			validateClusterOperatorsStable(clusterOperatorLister),
+		},
+		reconciliationChecks: []func(context.Context) (bool, error){
+			reconcileClusterOperatorsStable(clusterOperatorLister),
```
JoelSpeed: This makes me slightly nervous. If the cluster operators aren't stable, those instabilities will report up via those individual cluster operators. This implies to me that we might be duplicating the reporting, is that beneficial?

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	// 1. spec != status → a transition was requested, run reconcileTransition
+	// 2. spec == status, Progressing=True → transition applied, awaiting downstream reconciliation
+	// 3. spec == status, Progressing!=True → idle, ensure Upgradeable=True
+	if specTopology == "" || specTopology == statusTopology {
+		_, status, _, err := c.operatorClient.GetOperatorState()
+		if err != nil {
+			return err
+		}
+
+		if !v1helpers.IsOperatorConditionTrue(status.Conditions, transitionCondition) {
```
JoelSpeed: Why is the variable transitionCondition when  it is a progressing condition? Wouldn't it be more intuitive to call it progressingCondition?

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,306 @@
...
+		return err
+	}
+
+	specTopology := infra.Spec.ControlPlaneTopology
+	statusTopology := infra.Status.ControlPlaneTopology
+
+	// Three states:
+	// 1. spec != status → a transition was requested, run reconcileTransition
+	// 2. spec == status, Progressing=True → transition applied, awaiting downstream reconciliation
+	// 3. spec == status, Progressing!=True → idle, ensure Upgradeable=True
```
JoelSpeed, line 143: This implies to me we should see a big switch statement, the nested ifs right now represent this and perhaps it would be easier to read as a switch?

pkg/operator/topology_transition_controller/topology_transition_controller.go
JoelSpeed: Is there an `IsOperatorConditionFalse`? Rather than having to negate?

JoelSpeed: A controller that checks its own conditions to understand what state its in feels like an anti-pattern, you're creating a state machine.

Is there no way to tell what's happening from checking the state of the world around the operator and then setting the conditions based on observations?

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,306 @@
...
+	return c.reconcileTransition(ctx, syncCtx, infra)
+}
+
+// reconcileTransition finds the matching transition descriptor, runs preflight
+// validators, sets a progressing condition, and applies the status update.
+func (c *TopologyTransitionController) reconcileTransition(ctx context.Context, syncCtx factory.SyncContext, infra *configv1.Infrastructure) error {
+	transition, err := findTransition(infra, c.transitions)
+	if err != nil {
+		// Report via conditions, not sync error — these are user-fixable states
+		// that should not trigger WithSyncDegradedOnError.
```
JoelSpeed, line 200: Would we ever actually get there? The API should prevent these transitions no?

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,306 @@
...
+				Type:    transitionCondition,
+				Status:  operatorv1.ConditionFalse,
+				Reason:  "UnsupportedTransition",
+				Message: err.Error(),
+			}),
+			v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+				Type:    upgradeableCondition,
+				Status:  operatorv1.ConditionFalse,
+				Reason:  "UnsupportedTransition",
+				Message: "Cluster upgrade is not allowed while a topology transition is requested; revert spec.controlPlaneTopology to resolve",
```
JoelSpeed: This message would be more helpful if it told the user what the expected value was

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,306 @@
...
+		if !matchesSpec(transition.To, current.Spec) {
+			specChanged = true
+			return nil
+		}
+		transition.UpdateStatus(current)
+		_, err = c.infraClient.UpdateStatus(ctx, current, metav1.UpdateOptions{})
+		return err
+	}); err != nil {
+		return err
+	}
```
JoelSpeed: Rather than retry on conflict here, why not just return the error and re-reconcile and start the checks again with the new state?

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,306 @@
...
+func (c *TopologyTransitionController) checkClusterReconciliation(ctx context.Context) error {
+	_, status, _, err := c.operatorClient.GetOperatorState()
+	if err != nil {
+		return err
+	}
+
+	// Use the Progressing condition timestamp as the soak anchor. When
+	// Progressing is absent (safety-net path where the condition was lost),
+	// fall back to the Upgradeable condition — both are set at transition
+	// start, so either provides a valid lower bound.
```
JoelSpeed, line 301: And if neither exist?

pkg/operator/topology_transition_controller/transition_validation_checks.go
```diff
@@ -0,0 +1,274 @@
...
+const (
+	etcdEndpointsConfigMapName      = "etcd-endpoints"
+	etcdNamespace                   = "openshift-etcd"
+	etcdMembersAvailableCondition   = "EtcdMembersAvailable"
+	etcdMembersProgressingCondition = "EtcdMembersProgressing"
+)
+
+// validatePreflight runs global preflight checks followed by
+// transition-specific validators. Returns a combined error containing all
+// validation failures.
```
JoelSpeed, line 33: Is a combined error going to be readable? Are these ever likely to be transient errors or are they specific readable and actionable errors?

pkg/operator/topology_transition_controller/transition_validation_checks.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		return nil
+	}
+}
+
+// checkClusterOperatorsStable checks whether all ClusterOperators (except
+// config-operator itself) have reached a stable state. Returns a list of
+// descriptions for unstable operators (empty = all stable). This is the shared
+// core used by both the preflight validator and the post-transition
+// reconciliation check.
```
JoelSpeed, line 215: What happens if there are unrelated failures here?

jaypoulz:
@jeff-roche I have more specifics about ingress:
The issue is that the number of replicas is hard coded in the ingresscontroller:
```
[ec2-user@aws-jpoulin-dev ~]$ oc get ingresscontroller/default -n openshift-ingress-operator -o yaml
apiVersion: operator.openshift.io/v1
kind: IngressController
metadata:
  creationTimestamp: "2026-07-07T16:10:50Z"
  finalizers:
  - ingresscontroller.operator.openshift.io/finalizer-ingresscontroller
  generation: 1
  name: default
  namespace: openshift-ingress-operator
  resourceVersion: "53360"
  uid: f860ce4b-0e44-4878-ba92-f7e72d132094
spec:
  clientTLS:
    clientCA:
      name: ""
    clientCertificatePolicy: ""
  closedClientConnectionPolicy: Continue
  httpCompression: {}
  httpEmptyRequestsPolicy: Respond
  httpErrorCodePages:
    name: ""
  idleConnectionTerminationPolicy: Immediate
  replicas: 1
  tuningOptions:
    reloadInterval: 0s
  unsupportedConfigOverrides: null
```
We need to patch ingress operator to either detect drift or not hard code this.

go.mod
jeff-roche: Was accidentally updated when pulling in the new api, will revert to .0

pkg/operator/topology_transition_controller/types.go
jeff-roche: This is the definitive source for what are valid transitions and what gates them. When adding more you would expand the `TransitionDescriptor` array with your new values. Ex:

```go
return []TransitionDescriptor{
    // SNO to HA Compact on platformType: None
    {
        From: configv1.InfrastructureStatus{
            // ...
        },
        To: configv1.InfrastructureSpec{
           // ...
        },
        Validators: []TransitionValidatorFunc{
            // ...
        },
        UpdateStatus: func(infra *configv1.Infrastructure) {
             // ...
        },
    },
    // SNO to HA Compact on platformType: BareMetal
    {
        From: configv1.InfrastructureStatus{
            // ...
        },
        To: configv1.InfrastructureSpec{
           // ...
        },
        Validators: []TransitionValidatorFunc{
            // ...
        },
        UpdateStatus: func(infra *configv1.Infrastructure) {
             // ...
        },
    },
}
```

I'm pretty open to data structure changes here since I know this could be come unwieldy in the future but I haven't come up with a method of simplifying it yet.

pkg/operator/topology_transition_controller/topology_transition_controller.go
jeff-roche: Arbitrary number based on time estimates from a manual transition

pkg/operator/topology_transition_controller/topology_transition_controller.go
jeff-roche: It doesn't report the individual co status's, this primarily stops the transition status from flipping to done before the cluster has stabilized. It does give a nice direct data point for admins to look for if the transition is done or not

pkg/operator/topology_transition_controller/topology_transition_controller.go
jeff-roche: will update to `transitionProgressingCondition` to make it clearer

jeff-roche: This controller status is the source of truth for if a transition is in progress. Between reconciliation loops there is no good way to determine if a transition is in progress which is why I added the status field for the controller. I'm not sure of a better approach here but I'm open to suggestions

pkg/operator/topology_transition_controller/topology_transition_controller.go
jeff-roche: We can rely on the CEL expression in API but that expression is going to get massive if we expand the possible transitions. Having the controller validate the transitions currently is just a safety net we shouldn't ever hit, but if we want to drop the CEL expression at any point (because it became unwieldy), this would have us covered.

pkg/operator/topology_transition_controller/topology_transition_controller.go
jeff-roche: In that case I'm not sure we'd have a reference point to start the soak cycle from so it can just continue on

pkg/operator/topology_transition_controller/transition_validation_checks.go
jeff-roche: They should remain readable. Currently the only one is `reconcileClusterOperatorsStable`. These are defined at the controller level.

pkg/operator/topology_transition_controller/transition_validation_checks.go
jeff-roche: pre-transition we require all cluster operators to be stable. In this case transition will be blocked
post-transition we want to make sure all operators get back to a stable state. If they don't something went wrong with the transition and the offending operator will be reported

pkg/operator/topology_transition_controller/topology_transition_controller.go
jeff-roche: switched to this approach

pkg/operator/topology_transition_controller/topology_transition_controller.go
jeff-roche: Changed to a switch, I think it looks better but want to make sure you agree

pkg/operator/topology_transition_controller/types.go
```diff
@@ -0,0 +1,133 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	operatorv1listers "github.com/openshift/client-go/operator/listers/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+const (
+	// transitionAvailableConditionPrefix is used to build per-transition
+	// condition types: TopologyTransition_<Name>_Available
```
JoelSpeed: How does an end user understand what the expected conditions are if the types of the conditions are dynamic? Generally the condition types would be well defined and stable

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,303 @@
...
+	// Get the needed operator info to progress
+	_, status, _, err := c.operatorClient.GetOperatorState()
+	if err != nil {
+		return err
+	}
+	transitionProgressing := v1helpers.IsOperatorConditionTrue(status.Conditions, transitionProgressingCondition)
+	controllerUpgradeable := v1helpers.IsOperatorConditionTrue(status.Conditions, upgradeableCondition)
+
+	switch {
+	case specTopology != "" && specTopology != statusTopology:
```
JoelSpeed, line 155: Where's the case when spec == status and we aren't progressing? i.e. the steady state? Shouldn't that be setting appropriate conditions to their steady state values?

JoelSpeed, line 155: What happens if the transition is already progressing and the user tries to change the kind of transition that's occurring, is that blocked by the API currently?

pkg/operator/topology_transition_controller/transition_validation_checks.go
jaypoulz: client side we show that the the transition is not ready to proceed to but we allow the user to override this and force a transition by supplying an extra --allow-transition-with-warnings flag.

Backend side, the safest thing is is to not update the status field until these resolve and leave yourself in progressing state. We can probably degrade if we hit a timeout.

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,303 @@
...
+		// that should not trigger WithSyncDegradedOnError.
+		if _, _, condErr := v1helpers.UpdateStatus(ctx, c.operatorClient,
+			v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+				Type:    transitionProgressingCondition,
+				Status:  operatorv1.ConditionFalse,
+				Reason:  "UnsupportedTransition",
+				Message: err.Error(),
+			}),
+			v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+				Type:    upgradeableCondition,
```
JoelSpeed, line 209: Is this being aggregated into the main upgradeable condition by whichever loop would normally set that for config operator?

jeff-roche: Yes, it marks CCO us not upgradable which will block the cluster from being upgraded

pkg/operator/topology_transition_controller/topology_transition_controller.go
jeff-roche: If it's a valid transition (per the CEL expression), the api will allow it and CCO will try to reconcile it. Currently the only case that can happen is you trigger a transition but reset it back to SNO prior to the spec field being set. Otherwise if the Spec field has already been flipped, it's not a valid transition

jeff-roche: In that case, we return nil because there is nothing to be done. The other checks make sure if the conditions are not in their steady state values, they get evaluated and flipped if needed (`c.checkClusterReconciliation()`)

pkg/operator/topology_transition_controller/types.go
jeff-roche: This is from something I dropped from the PR, status reporting is likely to take a different form during the 5.1 cycle.

pkg/operator/topology_transition_controller/topology_transition_controller.go
```diff
@@ -0,0 +1,330 @@
...
+	controllerUpgradeable := v1helpers.IsOperatorConditionTrue(status.Conditions, upgradeableCondition)
+
+	switch {
+	case specTopology != "" && specTopology != statusTopology:
+		return c.reconcileTransition(ctx, syncCtx, infra)
+
+	case transitionProgressing:
+		return c.checkClusterReconciliation(ctx, infra)
+
+	case !controllerUpgradeable:
```
JoelSpeed, line 161: Does this also need to reset the progressing condition?

I'm trying to work out what happens when I request an invalid transition. I first fall into reconcile transition, that says upgradeable false and progressing false. Then I reconcile again and stay in reconcileTransition until something changes.

So as a user, I change the spec and go back to specTopology == statusTopology.

Now I fell into this case, at the moment it only resets one condition. The progressing condition would remain with the previous status no?

jeff-roche: In this case the `transitionProgressingCondition` would be `false` with reason `"UnsupportedTransition"`. You're saying reset the reason if specTopology == statusTopology and the last requested was Unsupported? I think that isn't currently covered here but potentially should be. WDYT?

jeff-roche:
~Added the following to the EP that needs to be reflected here:~ Done :heavy_check_mark: 
> the controller verifies an upgrade has not been triggered by CVO and then sets upgradable to false

pkg/operator/starter.go
```diff
@@ -264,6 +267,41 @@ func (o *OperatorOptions) RunOperator(ctx context.Context, controllerContext *co
...
+			operatorInformers.Operator().V1().Etcds().Informer(),
+			configInformers.Config().V1().ClusterOperators().Lister(),
+			configInformers.Config().V1().ClusterOperators().Informer(),
+			clock.RealClock{},
+			controllerContext.EventRecorder,
+		)
+
+		go operatorInformers.Start(ctx.Done())
+		go kubeInformersForEtcd.Start(ctx.Done())
+		go topologyTransitionController.Run(ctx, 1)
```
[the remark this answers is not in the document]
JoelSpeed: This looks like it needs attention

